PR #56, opened 2014-08-15
Get the html source of messages from html_codesniffer

mfairchild365:
It is sometimes helpful to see the html source of a message returned by html_codensiffer.

I made these changes while trying to locate the source of a message in pa11y.  The message was present in pa11y, but wasn't being reported by html_codesniffer bookmarklet when I ran it in the browser, so I had no idea what the context of the error was.  (it turns out it was because phantomjs thinks it is a touch browser)

PR #66, opened 2015-03-17
Get and truncate the html source of messages from html_codesniffer

atambo:
This is basically a rebase of #56 against master plus adding some basic truncation.

### DIFF
--- a/lib/sniff/handle-result.js
+++ b/lib/sniff/handle-result.js
@@ -42,7 +42,8 @@ exports.sanitizeMessages = function (messages) {
 		return {
 			code: message.code,
 			message: message.msg,
-			type: messageTypeMap[message.type] || 'unknown'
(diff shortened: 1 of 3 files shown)


atambo:
@rowanmanning, @mfairchild365: Here is the rebase & truncation.

Pikaling:
Glad to see you're working on this. I find pa11y difficult to use at the moment because I don't know where the element is that's causing the error.

rowanmanning:
Hi @atambo, thanks for this. I've merged into the `develop` branch and made a small change I'd like to run by you before tagging for release: https://github.com/nature/pa11y/commit/413141d40cd993d33600d3dce17b0872092e6bce

atambo:
@rowanmanning, good catch. Now the max length of html is 251 which sounds like a sane amount to me.

Issue #77, opened 2015-06-08
Add element XPath to result


rowanmanning:
See discussion in #61. This could be done quite easily in pa11y 2.0, you'd need to get each element's XPath in [lib/inject](https://github.com/nature/pa11y/blob/master/lib/inject.js) (which run's within PhantomJS). You could use one of the functions in these SO answers:

  - [How to calculate the XPath position of an element using Javascript?](http://stackoverflow.com/questions/3454526/how-to-calculate-the-xpath-position-of-an-element-using-javascript)
  - [Javascript get XPath of a node](http://stackoverflow.com/questions/2661818/javascript-get-xpath-of-a-node)

rowanmanning:
@adelevie, opening a new issue for discussions around xpath. This would be really useful

adelevie:
Potentially dumb question, but how does HTML Code Sniffer know where to place the tooltips when running a sniff?

rowanmanning:
Each message in HTML CodeSniffer has a reference to a DOM element so I guess it just calculates where to place the tooltip based on it's dimensions and position. I did a bit of investigation into capturing screenshots of these with pa11y but it's a considerable amount of work

adelevie:
I'd be less interested in screenshots, but would love to get the XPaths or CSS selectors of these elements from `pa11y`. The end goal is a "flat" HTML page of the rendered test page with the tool tips. I want to make reports linkable, rather than asking folks to install and run the HTML CS bookmarklet.

adelevie:
See https://github.com/18F/pulse/issues/195 for some context of my goals.

rowanmanning:
Ah I see, well I know it'd also be really useful for pa11y-dashboard on our side, so I'll definitely get time to implement something. Do you have a preference for XPath/CSS? If they're simple enough to implement I could add both or make it configurable.

adelevie:
I don't have a preference--my only goal is to be empowered to create my own HTML tooltips for "static" reports. Whichever element identifier makes that happen is what I want. Let me know if/how I can help with this.

adelevie:
Here's a (horrible) proof of concept for CSS selectors. In `inject.js`:

```javascript
  function fullPath(el){
    var names = [];
    while (el.parentNode){
      if (el.id){
        names.unshift('#'+el.id);
        break;
      }else{
        if (el==el.ownerDocument.documentElement) names.unshift(el.tagName);
        else{
[...]

PR #78, opened 2015-06-09
CSS Selectors

rowanmanning:
Add the element's CSS selector to the results, completing #77 (css selectors were easier than XPath)


adelevie:
From skimming the code everything looks good to me. I'll download and test it out later today. Thanks so much for adding this.

PR #81, opened 2015-06-16
Allow specificing the HTML_CodeSniffer Library location

mfairchild365:
We are using a customized fork of the HTML_CodeSniffer Library, so being forced to use the included one is a big step backwards for us.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Usage: pa11y [options] <url>
     -p, --port <port>          the port to run PhantomJS on
     -t, --timeout <ms>         the timeout in milliseconds
     -d, --debug                output debug messages
+    -h, --htmlcs <url/path>    The url or path to the HTML_CodeSniffer Library
 ```
 
 ### Running Tests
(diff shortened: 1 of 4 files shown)


rowanmanning:
Thanks @mfairchild365 :) are you super attached to specifying a URL for this? Just wondering if it can be simplified a little. Fine if you need URL support, not just local file-system. Also not keen on using 'local' as a default – we could simplify the logic and just set the default to be `__dirname + '/vendor/HTMLCS.js'`

rowanmanning:
Actually maybe it'd be simpler if we just tested whether the `htmlcs` option matches `/^https?:\/\//`? Let me know if you want me to do this – I have some time today/tomorrow.

mfairchild365:
@rowanmanning The regex test would probably be simpler.  I'm not super familiar with phantomjs, but I was assuming that you could specify a relative URL to include, and therefor the regex wouldn't work.  I don't think that I will have time to work on this today, so you are more than welcome to. :)

mfairchild365:
Also, specifying a URL would be preferred.  :)

PR #82, opened 2015-06-16
Wait a little bit to let the page finish loading before testing

mfairchild365:
I'm not sure what the best way to handle this might be.

We have a case where a javascript widget injects some CSS to style otherwise inaccessible content (contrast ratio).  However, the javascript widget injects the CSS just after page load, so when pa11y tests the page, the content is flagged as an error.  If you run the html_codesniffer bookmarket, everything is okay because the CSS had had a chance to be injected.

The point is, we don't know when JS will inject things make make things accessible or inaccessible.  The hope here is to add a small buffer to handle the majority of edge cases.  

Ideas and feedback is very welcome.  :)

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -126,7 +126,11 @@ function testPage (options, browser, page, done) {
 				if (typeof window.callPhantom !== 'function') {
 					return new Error('Pa11y could not report back to PhantomJS');
 				}
-				injectPa11y(window, options, window.callPhantom);
+
+				window.setTimeout(function () {
+					//Wait a bit to let the page finish loading
+					injectPa11y(window, options, window.callPhantom);
+				}, 500);
 			}, next, {
 				ignore: options.ignore,
 				standard: options.standard


rowanmanning:
There are a couple of ways we could do this, I guess. I'll have a proper think, might come up with more later.

1. Allow specifying a wait time between page load and the tests running (not sure I like this, it feels brittle)
2. Allow running a function before pa11y is injected that tests whether all your code has run somehow
3. Allow ignoring of elements based on their HTML or selector (probably not the best solution but could be useful in other cases for ignoring things like adverts)

mfairchild365:
Great ideas, but I don't think 2 and 3 will work in our case.  We are auditing hundreds of unique sites on our campus which have little knowledge of or control over.  A broad and generalized solution would be best.

1. You are right, it does feel brittle, but I can't think of a better way.
2. That will work as long as you are auditing only your own code, which we are not.
3. Same as 2, that could work as long as you are only auditing your own code.

mfairchild365:
@rowanmanning Have you thought any more about this?  I could add a configuration option to allow specifying the amount of time to wait before running the tests, and have it default to 0.

rowanmanning:
Yep, I've been thinking about it and maybe for now a wait option is enough. I've been too busy today, but intending on getting it done this week. Does that work for you?

mfairchild365:
that sounds great @rowanmanning :)

PR #84, opened 2015-06-23
Add an option which makes pa11y wait for a configured amount of time before running tests

rowanmanning:
This allows for testing pages after client-side JavaScript has been run. It's not bullet-proof but works for now. Later we may enable execution of arbitrary scripts before the HTML CodeSniffer run which will be a more solid way to do this. Addresses #82.

@mfairchild365, happy with the implementation? Should it solve your issue?


debugwand:
looks good

mfairchild365:
Looks good! Thanks for your help with this @rowanmanning

rowanmanning:
No problem :) glad I can help make pa11y more useful for you

PR #97, opened 2015-10-07
Pa11y 3.0

rowanmanning:
See the [Truffler 2.0 PR](https://github.com/nature/truffler/pull/5) and the [README](https://github.com/nature/pa11y/blob/3.x/README.md) for information on what pa11y 3.0 changes.

Before:
```js
pa11y({
    // ...
}, function (error, test, exit) {
    test('http://www.nature.com/', function (error, results) {
        console.log(results);
        exit();
    });
});
```

After:
```js
var test = pa11y();

test.run('http://www.nature.com/', function (error, results) {
    console.log(results);
});
```

The command-line interface hasn't changed.

Still to-do:

- [x] Get Truffler 2.0 released properly and update `package.json`
- [x] Work out how to handle the jsdom-based tests
- [ ] Consider re-licensing under the more permissive LGPL-3.0 to match Shunter


rowanmanning:
Test failures are all due to a drop in coverage. The coverage has actually _increased_ in 3.0, but I've had to comment out jsdom-related tests temporarily so coverage reporting is way off. Will fix before merge.

lib/inject.js
```diff
@@ -85,10 +85,16 @@ function injectPa11y(window, options, done) {
 	}
 
 	function buildElementIdentifier(element) {
-		var identifier = (element.id ? '#' + element.id : element.tagName.toLowerCase());
+		if (element.id) {
+			return '#' + element.id;
```
moddular: return identifier; ?

rowanmanning: That line's been removed :) the new conditional is so that we don't need the ternary and we return early

moddular: Too early for proper comprehension apparently :-/

moddular:
:+1:

PR #105, opened 2015-11-20
Add parameter for phantomjs.exe path. Fixes #102

### LINKED ISSUE #102: Add a PhantomJS path option/parameter
At present, pa11y required PhantomJS to be installed globally. In order to automate the full installation, would it be possible to [add PhantomJS as a dependency](https://github.com/addyosmani/a11y/blob/master/package.json#L59)?

If not, an alternative would be to add a parameter for the path to the binary.

carbonrobot:


rowanmanning:
I haven't checked properly (I can do today) but we may need to make a change to Truffler for this to work. This line here checks that `phantomjs` is available in `$PATH`, so the library will fail with an error anyway https://github.com/nature/truffler/blob/master/lib/truffler.js#L41

rowanmanning:
Hi @carbonrobot, sorry for the long wait time on this. Yes your example should be all that's required in Truffler I think. Would you like to open a PR? If not then I can do so in the next day or two :)

rowanmanning:
OK Truffler `2.1.0` has been released. You'll need to bump the version in package.json to `~2.1` then this is ready to merge.

rowanmanning:
Thanks for your work @carbonrobot :) I've merged and released 3.1.0, you should be able to install the new version now

PR #109, opened 2015-12-16
Compute several URL (received from STDIN)

### LINKED ISSUE #91: Ability to test multiple URLs in a single invocation of the CLI
pa11y is fantastic, thanks for all the hard work that went into!

Having the ability to provide 2+ URLs on the CLI would be helpful.

Hywan:
Fix #91.

Now we can do:

    $ echo 'file:///tmp/a.html\nfile:///tmp/b.html' | pa11y -0

to run pa11y on `a.html` and `b.html`.

We can use it with a crawler so:

    $ crawl http://example.com/ | pa11y -0

Note: URLs are synchronously tested.

Thoughts?

### DIFF
--- a/bin/pa11y
+++ b/bin/pa11y
@@ -173,7 +173,7 @@ function runProgram(program) {
 		);
 	} else if (program.stdin) {
 		var urls = getRunQueue(test, options);
-		readline.createInterface(process.stdin, process.stdout).on('line', function(url) {
+		readline.createInterface(process.stdin, undefined).on('line', function(url) {
 			urls.push({url: url});
 		});
 	} else {
--- a/bin/pa11y
+++ b/bin/pa11y
@@ -22,13 +22,16 @@ var path = require('path');
 var pkg = require('../package.json');
 var program = require('commander');
 var pa11y = require('../lib/pa11y');
+var readline = require('readline');
+var async = require('async');
+var chalk = require('chalk');
 
 configureProgram(program);
 runProgram(program);
 
 function configureProgram(program) {
 	program.version(pkg.version)
-		.usage('[options] <url>')
+		.usage('[options] [url]')
 		.option(
 			'-s, --standard <name>',
 			'the accessibility standard to use: Section508, WCAG2A, WCAG2AA (default), WCAG2AAA'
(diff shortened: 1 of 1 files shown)


Hywan:
Note: This might not be the best way to do it, I just discovered your project yesterday.

Hywan:
It seems like there is an issue with concurrent accesses to I-don't-know-what. I have this random error:
```
events.js:141
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at Object.exports._errnoException (util.js:874:11)
    at exports._exceptionWithHostPort (util.js:897:20)
    at WriteWrap.afterWrite (net.js:763:14)
```

Hywan:
Fixed. New patches are coming :-).

Hywan:
And now it's green :-).

rowanmanning:
Hey @Hywan, I've got a couple of days off and won't have chance to look at this immediately. Thanks for your work! I'll review once I'm back in the office next week

Hywan:
@rowanmanning Excellent :-). Another PR is probably coming with multiple output support for reporter (e.g. html).

Hywan:
I have rebased on `master` —since 3.2.0— and I have added the `-o/--output` options. This can be included in the same PR I guess, this is strictly related.

Hywan:
I am just quoting the commit message of the `-o/--output` commit:

Because we can compute several URL (from STDIN), the reporter may output
their results in different files. The `-o/--output` option takes a
valid filename as value. The given filename will receive the output from
the reporter (only the output, not the error). This is also useful when
dealing with a single URL (it can replace the `>` Shell redirection).

This option understands a special keyword: `%HASH%`, which represents
the SHA1 of the currently computed URL. So for instance:

    $ echo file:///a.html\nfile:///b.html | \
          pa11y -0 -r html -o 'output/%HASH%.html'

will produce 2 HTML reports in the `output/` directory.

To produce an index, simply insert a command between the pipe for
instance.

rowanmanning:
OK, I should _finally_ have time to look at this properly now. I had to push a bug fix yesterday so I'm afraid you'll have to rebase again first though. Thanks for your patience :)

Hywan:
@rowanmanning It's fine, that's the game ;-).
Now it's rebased :-). I did try to replicate your bug fix. Please, check!

bin/pa11y
```diff
@@ -81,29 +90,28 @@ function configureProgram(program) {
...
 			if (error) {
 				options.log.error(error.stack);
-				process.exit(1);
+				return callbacks.error(error);
 			}
 			if (reportShouldFail(program.level, results, program.threshold)) {
-				process.once('exit', function() {
-					process.exit(2);
-				});
+				return callbacks.error(error);
```
rowanmanning: This return prevents the results from ever being output. If we add the following above then it works and the tests pass again:

```
options.log.results(results, url);
```

rowanmanning:
It'd be good to get a couple of integration tests in place for this, I can help if you need it, but I can imagine accidentally breaking this myself at some point!

bin/pa11y
Hywan: So you suggest to only remove the `return` statement?

Hywan:
@rowanmanning Do you have already an integration test suite?

bin/pa11y
rowanmanning: I don't think so, because otherwise it will `process.exit` with `0` always. I think. Haven't tested that

rowanmanning:
Yep, they're in `test/integration` but will probably need some work – they currently assume that you're calling pa11y with no piped input etc. Like I said, I'm happy to work on them if you don't have the time :)

Hywan:
@rowanmanning I have to admit that I would greatly accept your help on the integration test suite ;-)!

bin/pa11y
Hywan: Hmm, we can call `options.log.results(results, url);` before the return too.

rowanmanning:
I'll try and find some time today or over the weekend :)

bin/pa11y
rowanmanning: Don't see why not!

Hywan: Done.

Hywan:
ping :-)?

rowanmanning:
Hi Ivan, after some discussion in our latest Pa11y meeting we've decided that multiple URL testing is probably out of scope for the main library (in this major version at least).

Sorry, I realise you've put time and effort into this and I should have got back to you sooner. If you're looking for a multiple URL tester then 18F are working on a [Lambda function](https://github.com/18F/pa11y-lambda), and I'm also going to be adding more examples in #153.

I hate this part of being an OSS maintainer, and I hope you're not dissuaded from contributing to Pa11y again in future – we'd be more than happy to have your input.

Hywan:
It's OK :smiley:. You are very correct, not rude, everything is right. We have our own implementation for handling many URLs in [`liip/TheA11yMachine`](https://github.com/liip/TheA11yMachine/). This PR was an extract of our implementation at that time. Feel free to watch.

Issue #98, opened 2015-10-08
Allow running of additional PhantomJS code before pa11y runs


rowanmanning:
Currently there are some issues around login forms and pages that rely on a lot of JavaScript to render. We have the `wait` option for the latter, but it's not ideal.

Proposing adding the ability to run additional custom PhantomJS code before pa11y to prepare the page for testing. Not sure how this would look yet, but I'll throw together some ideas when I have time.

GlynnPhillips:
I started looking at this but then realised my solution was not the best for all cases so it might be helpful to figure out what the user cases we are trying to cover. A couple I can think of:

* Allowing for UI changes e.g. clicking on an element to open a hidden section
* Adding data and submitting form elements e.g. Log in
* Allowing for ajax requests to be made to update content

Some of the problems I was hitting when thinking about these was how to pass back a message to pa11y that the script had finished and the testing could continue. Phantom has a couple options which could work like `onCallback` and `onConsoleMessage` but I couldn't make my mind up as to which would be right in this case. 

It also might be worth defining if these interactions should/would allow for page refreshes or location changes so this might effect the way in which the message is passed back that the script has finished, or might not be possible in cases.

JeffHall:
I've just started experimenting with pa11y, and I'm very impressed by what I've seen so far!

So the obvious caveat is that I'm a newbie, but in thinking about this issue and the user cases that @GlynnPhillips mentions above, these are the ones that jump out at me:

* Allowing for UI changes e.g. clicking on an element to open a hidden section
* Allowing for ajax requests to be made to update content

The AngularJS app that I'm working with has several *ngClick* directives that change the state of the page display (essentially show/hide behaviors), so using PhantomJS to navigate to the expected page "state" would be ideal before the pa11y verification itself is run.

Now my challenge is to learn enough JavaScript and Node so that I can actually contribute to this great project!

rowanmanning:
After a bit of thought I think that the simplest way to do this would be to give the end user a place to hook into PhantomJS directly. We could just stick an [extra function in here](https://github.com/nature/pa11y/blob/master/lib/pa11y.js#L123) which just runs the user code if there is some, then they can do whatever the hell they want and we cater for all the above use-cases (and more that we can't think of).

API-wise it'd be nice to do something like:

### Command Line

```sh
pa11y --extras ./my-function.js nature.com
```

Where `./my-function.js` is a Node.js module that exports a single function (see below for the expected args for that function).

### JavaScript

```js
pa11y({
[...]

PR #111, opened 2015-12-19
Adding an option for running JS before testing

GlynnPhillips:
DO NOT MERGE
I wanted to get this PR in early for discussion but its not complete yet and ready for merging.

This is an attempt to add the new functionality suggested in issue #98.

This will add the `--injectJs` option to pass in a file path to a
JavaScript file to be run pa11y is run.

That JavaScript file has to `console.log('pa11y: script complete')`
before pa11y will be run.

I am not 100% sure how I feel about this `console.log` approach but it
felt like it kept the supplied JS simple and without any phantomjs
knowledge required compared to the callback approach.

DO NOT MERGE
Tests still need to be written.

### DIFF
--- a/bin/pa11y
+++ b/bin/pa11y
@@ -70,6 +70,10 @@ function configureProgram(program) {
 			'the time to wait before running tests in milliseconds'
 		)
 		.option(
+			'-I, --injectJs <path>',
+			'a script that can be injected into phantomjs before pa11y runs'
+		)
+		.option(
 			'-d, --debug',
 			'output debug messages'
 		)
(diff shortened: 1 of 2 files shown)


bin/pa11y
```diff
@@ -70,6 +70,10 @@ function configureProgram(program) {
 			'the time to wait before running tests in milliseconds'
 		)
 		.option(
+			'-j, --javascript <script>',
```
GlynnPhillips: I'm not sure on the naming of this option so just went with this for now, better suggestions are very welcome.

GlynnPhillips:
Actually now I understand the workings of this more I think it needs more work, will reopen a PR whent his has been done

PR #115, opened 2016-01-11
Multiple ignore options

tavvy:
use commander coercion to allow repeatable ignore options


rowanmanning:
Looks great :) one small addition needed: the README contains a copy of the command line help, so the `--ignore` docs need updating https://github.com/nature/pa11y#command-line-interface

tavvy:
ah ofcourse. do you think this makes sense?
"types and codes of messages to ignore, a repeatable value or separated by semi-colons"

PR #124, opened 2016-01-29
Allow more standards. Do not hardcode the list

Hywan:
This is necessary when we add our own standards and would like to use them. This is not a feature, this is kind of a bug fix :-).

Cheers!

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -47,13 +47,14 @@ module.exports.defaults = {
 		}
 	},
 	standard: 'WCAG2AA',
+	allowedStandards: ['Section508', 'WCAG2A', 'WCAG2AA', 'WCAG2AAA'],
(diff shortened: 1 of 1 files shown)


rowanmanning:
Looks great, makes a lot of sense. One thing I'm not sure on is the naming of the new option. What do you think of `allowedStandards` instead of `standards`, just to be explicit?

Hywan:
@rowanmanning Agree.

Hywan:
@rowanmanning Rebased and pushed!

Hywan:
Why is it failing?

rowanmanning:
It's not failing, it's just Coveralls being weird:

> Coverage decreased (-0.0%) to 99.167%

Thanks!

Hywan:
\o/

Issue #133, opened 2016-02-09
Ignore everything but a given selector


GlynnPhillips:
Previously raised by @davidjb in https://github.com/springernature/pa11y/issues/108

> +1. In my case I'm interested in also ignoring everything but a specific selector, so perhaps include/exclude options would satisfy both. Having pa11y test the whole page but just filter the report output could be a solution.

> In my specific use case, I have documentation pages from an upstream web framework and want to test a specific selector on each page (eg excluding everything else).

One way of achieving this could be to altering the second option [here](https://github.com/nature/pa11y/blob/master/lib/inject.js#L16) to use a query selector instead of the whole document

GlynnPhillips:
@nickcall and I have looked at this today and found a slightly problem with the proposed approach.

By altering the second option [here](https://github.com/nature/pa11y/blob/master/lib/inject.js#L16) to select the area of the page to test you could end up with false errors in certain examples.

One example would be using anchors to jump to another region on the page. If the anchor as within the testing area but the target element is outside of the testing area Pa11y will still return an area that the named anchor does not exist.

The alternative approach would be to filter the results after the full page has been tested however this would be considerably more work and the filtering would be an expensive process.

@rowanmanning what do you think?

GlynnPhillips:
One approach we have come up with to achieve this without too much complexity would require pa11y to dictate that the element we want to limit our results too must have a data attribute applied to it. This data attribute could be named anything the user wants and would be passed in as an option. 

Quick example of how `/lib/inject.js` could look

Inside the [buildElementIdentifier function](https://github.com/springernature/pa11y/blob/master/lib/inject.js#L72)

```js
function buildElementIdentifier(element) {
    if(element.getAttribute(options.testArea)) {
         return options.testArea;
    }
    // The rest of this function would stay the same
}
```

So if `options.testArea` equalled `data-test-region` the `message.selector` would look something like this `html > body > data-test-region`

The [isMessageWanted functuion](https://github.com/springernature/pa11y/blob/master/lib/inject.js#L103) could then easily filter out unwanted messages without at any point in this process needing to keep all the properties of an element to filter on.

A couple of gut feelings I have about this approach is that it might not be overly elegant and also we would be enforcing the use of data attributes (or a similar attribute)

rowanmanning:
Alternatively you could allow any selector like we discussed before but still run the test on `document`. Then after the test has run you'd have to loop over the results and exclude any relating to elements that aren't children of the specified element. Does that make sense?

[...]

PR #136, opened 2016-02-12
Adds the ability to hide elements from pa11y testing

GlynnPhillips:
This will allow areas of the page to be hidden before running HTML code
sniffer against the page. This relies on the fact that HTML code sniffer
will ignore anything thats `visibility` style is set to `hidden`.

The option accepts a string which allows for coma seperated selectors in
the same way as `querySelectorAll`.

e.g. "#modal, .adverts, div[data-hide]"

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,20 +107,21 @@ Usage: pa11y [options] <url>
 
   Options:
 
-    -h, --help                 output usage information
-    -V, --version              output the version number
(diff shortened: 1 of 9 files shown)


rowanmanning:
Looks great! Are we planning to merge alongside #133 once we have a PR for it?

GlynnPhillips:
Yeah @nickcall and myself are planning to look at #133 this week

PR #138, opened 2016-02-26
Adds ability to define testing area

GlynnPhillips:
Allows you to test a part of the page and ignore everything outside of it.

I wanted to raise this PR to get some feedback on the naming before tidying up and putting in the tests.

I was struggling with the option name and I am not sure how people feel about `selector` ?

### DIFF
--- a/lib/inject.js
+++ b/lib/inject.js
@@ -40,7 +40,11 @@ function injectPa11y(window, options, done) {
 	}
 
 	function processMessages(messages) {
-		return messages.filter(isMessageInTestArea).map(processMessage).filter(isMessageWanted);
+		if (options.rootElement) {
+			messages = messages.filter(isMessageInTestArea);
+		}
+
+		return messages.map(processMessage).filter(isMessageWanted);
 	}
 
 	function processMessage(message) {
--- a/lib/inject.js
+++ b/lib/inject.js
@@ -115,9 +115,9 @@ function injectPa11y(window, options, done) {
 	}
 
 	function isMessageInTestArea(message) {
-		var testArea = document.querySelector(options.selector);
+		var testArea = window.document.querySelector(options.selector);
 
-		if (testArea) {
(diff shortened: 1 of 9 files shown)


rowanmanning:
I like it! Not sure about the name `selector` either, but struggling to think of an alternative right now. Maybe something like `rootElement`?

GlynnPhillips:
Ok I will work on this over the weekend and we can have a think and change it once I have everything worked out and tests written.

rowanmanning:
Cool :) don't work too hard!

GlynnPhillips:
I have now added some tests and documentation. Happy for anyone to correct my documentation if they have a better way of explaining the feature.

GlynnPhillips:
Thanks for the help with documentation @rowanmanning I have made those changes

lib/inject.js
```diff
@@ -26,7 +26,7 @@ function injectPa11y(window, options, done) {
 	}
 
 	function processMessages(messages) {
-		return messages.map(processMessage).filter(isMessageWanted);
+		return messages.filter(isMessageInTestArea).map(processMessage).filter(isMessageWanted);
```
rowanmanning: Oh actually one more small tweak. Currently if there is no `rootElement` specified, we run `querySelector` on `null`. I wonder whether this function would be more optimal if it was:

```js
function processMessages(messages) {
	if (options.rootElement) {
		messages = messages.filter(isMessageInTestArea);
	}
	return messages.map(processMessage).filter(isMessageWanted);
}
```

rowanmanning:
Sorry, should have spotted that last comment earlier. Otherwise really happy with this :) feel free to merge this and #136 tomorrow and release as `3.6.0`!

lib/inject.js
GlynnPhillips: DOH! yeah I moved that when debugging and forgot to replace it. Good catch

Issue #146, opened 2016-05-02
CLI hide-elements option not working as expected


davidmlentz:
From command line, including the `hide-elements` option appears to have no effect. Below is some output where the error listed should have been ignored:

**Command:** `pa11y --ignore "warning;notice" --hide-elements ".hidden_link" https://oit.boisestate.edu/`

**Result:**
    Error: Anchor element found with a valid href attribute, but no link content has been supplied.
    ├── `WCAG2AA.Principle4.Guideline4_1.4_1_2.H91.A.NoContent`
    ├── `#footer > div:nth-child(3) > a`
    └── `<a class="hidden_link" href="http://go.boisestate.edu/" target="_blank" style="visibility: hidden;"></a>`

rowanmanning:
Hi @davidmlentz, thanks for reporting this. I've verified it with a breaking test. @GlynnPhillips, do you have any thoughts?

GlynnPhillips:
Thanks @davidmlentz for the detailed report. This functionality relies on functionality of HTML_codesniffer and I can see in #143 that this was updated last week. Tomorrow I will look to see if this could be the cause and if not what the cause is.

GlynnPhillips:
I have only managed to take a quick look at this but it looks like making an element `visibility:hidden` will cause `HTML_codesniffer` to not report some errors but will report others which was a bit of an oversight on my part.

So I think this means we need to rethink this functionality within `pa11y` as we cant rely on `HTML_codesniffer` to not report errors when an element is hidden which now that I think about it makes sense for some errors.

I will take a look at `HTML_codesniffer` in some more detail to see if there is any way we can use functionality it already has else we will need to filter out errors within `pa11y`.

GlynnPhillips:
Hi @davidmlentz sorry for the delay on this, I have started on a fix for this which you can see here: https://github.com/springernature/pa11y/commit/fecadb960eadf86a832fdc0716e31f87a571ccfe

Although this is now working the code needs some tidying up to make it nicer to understand and we need to make sure this time that I added plenty of tests to proof it works.

Hopefully I should get some time over the weekend to do both of these things so we can push a fix next week.

GlynnPhillips:
Hi @davidmlentz 

Sorry about the delay on this when I initially thought I would be done in a couple days I hit some complications thrown up by Phantom. However I am now pretty much done with this and you can see the changes in #160.


I will add a couple more tests to make sure this time we are not pushing faulty functionality and then we will release this.

[...]

PR #147, opened 2016-05-02
Write a test to prove that hide elements is broken

rowanmanning:
See #146 for the original bug report. It looks like our current method for hiding these elements (adding `visibility:hidden`) only works for certain rules. We need to rethink the way this is implemented.


rowanmanning:
@GlynnPhillips did you say that you copied this test into your branch? If so, I can close

GlynnPhillips:
Yep its here, so will close 

https://github.com/pa11y/pa11y/blob/broken-hide-elements-test/test/integration/mock/html/hide-elements.html#L11

Issue #144, opened 2016-04-25
Document that `make build-htmlcs` won’t run with npm 3.x


joseluisbolos:
The command `make build-htmlcs` that is used to manually pull the latest version of HTML codesniffer  won’t run with npm 3.x at the moment, only with npm 2.x.

We need to document that, at least until it can be fixed.

rowanmanning:
I'm pretty sure this has magically fixed itself, because I can run with npm `3.10.10` fine. Let's close, as it's only ever one of the core maintainers who would do this work anyway.

PR #151, opened 2016-05-18
switch to https git clone, and make some corrections to make ops due …

hollsk:
…to new HTMLCS structure

I've switched the git clone from ssh to https to correct the problem that Jose refers to in #144 - where npm has introduced a breaking change that stops ssh cloning from working properly. There's a lengthy thread about that here: https://github.com/npm/npm/issues/8031

After I did this, I realised that HTML codesniffer has some recent lib changes, so I've updated relevant locations for a couple of tasks. `make build-htmlcs` should now work correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,11 +47,11 @@ test-integration:
 # Rebuild a local HTML CodeSniffer
 build-htmlcs:
 	@echo "$(C_CYAN)> building HTML CodeSniffer$(C_RESET)"
-	@npm install git+ssh://git@github.com:squizlabs/HTML_CodeSniffer.git
+	@npm install git+https://git@github.com/squizlabs/HTML_CodeSniffer.git
 	@npm install -g grunt-cli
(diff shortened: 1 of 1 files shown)


joseluisbolos:
Looks good! 👍

rowanmanning:
I'm still having some trouble with npm 3. I think we might need to test cloning HTML CodeSniffer and running their build by itself with npm 3 before we can complete this.

hollsk:
@rowanmanning what were the problems you were having? I'd like to either merge this or abandon the branch properly, but I'm not sure what I'm looking for r.e. npm 3 😁

rowanmanning:
I don't remember 😁 will have another look ASAP and describe the issues properly

rowanmanning:
As mentioned in #144, I don't think this is an issue any more. We can reopen/investigate if it is, but tbh we're the only people who run this anyway :)

PR #158, opened 2016-05-24
Make the built-in reporters usable in JavaScript

### LINKED ISSUE #137: Allow Reporter Options for JavaScript Interface
Would like to use the reporter options in the Javascript Interface.  The raw JSON results are great but would like to be able have the html option.

rowanmanning:
This makes the built-in reporters usable with the JavaScript API. This should resolve #137.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ The following locations will be checked:
 <cwd>/rainbows
 ```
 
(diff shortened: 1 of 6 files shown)


howlingmad:
code looks good, and the explanation in the Readme makes sense to me.

PR #160, opened 2016-05-25
Fixes the hide-elements options

GlynnPhillips:
The `hide-elements` options was broken due to a misunderstanding of how
`HTML_codeSniffer` worked.

This change moves away from relying on `HTML_codeSniffer` for this
feature and instead filters out errors that are unwanted.

### DIFF
--- a/lib/inject.js
+++ b/lib/inject.js
@@ -8,22 +8,8 @@ function injectPa11y(window, options, done) {
 		2: 'warning',
 		3: 'notice'
 	};
-	if (options.hideElements) {
-		hideElements();
-	}
 	setTimeout(runCodeSniffer, options.wait);
 
-	function hideElements() {
-		try {
-			var elementsToHide = window.document.querySelectorAll(options.hideElements);
-			for (var i = 0; i < elementsToHide.length; i++) {
-				elementsToHide[i].style.visibility = 'hidden';
-			}
-
-		} catch (error) {
-			reportError('Hiding elements: ' + error.message);
-		}
-	}
 
 	function runCodeSniffer() {
 		try {
@@ -42,6 +28,10 @@ function injectPa11y(window, options, done) {
 	function processMessages(messages) {
 		if (options.rootElement) {
 			messages = messages.filter(isMessageInTestArea);
+		}
+
+		if (options.hideElements) {
+			messages = messages.filter(isElementOutsideHiddenArea);
 		}
 
 		return messages.map(processMessage).filter(isMessageWanted);
@@ -128,6 +118,16 @@ function injectPa11y(window, options, done) {
 		}
 	}
 
+	function isElementOutsideHiddenArea(message) {
+		var hiddenSelectors = options.hideElements.split(',');
+		var element = message.element;
+		var elementsWithinHiddenSelectors = hiddenSelectors.filter(function(selector) {
+			return hiddenAreasContainsElement(element, selector);
+		});
+
+		return elementsWithinHiddenSelectors.length ? false : true;
+	}
+
 	function isElementWithinTestArea(child, parent) {
 		var node = child.parentNode;
 		while (node !== null) {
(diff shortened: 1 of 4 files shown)


lib/inject.js
```diff
@@ -140,6 +138,25 @@ function injectPa11y(window, options, done) {
 		return false;
 	}
 
+	function hiddenAreasContainsElement(element, hiddenSelectors) {
+		var hiddenElements = window.document.querySelectorAll(hiddenSelectors);
+
+		for(var i = 0; i < hiddenElements.length; i++) {
+			if(element.isEqualNode(hiddenElements[i])) {
```
GlynnPhillips: I would have preferred to use `Element.matches` but unfortunately its not supported within Phantom and polyfilling it would have meant faking a lot of the DOM in the unit tests so I have decided to go with `isEqualNode` instead. 

This unfortunately this means I have had to use `querySelectorAll` and loop over every node it matches.

rowanmanning: If it's fast enough with most sites then I think it's OK :) we can always refactor if performance is an issue.

test/unit/lib/inject.js
```diff
@@ -30,29 +30,116 @@ describe('lib/inject', function() {
...
+			{
+				outerHTML: '<element class="hide">parent</element>'
+			}
+		]);
+		window.HTMLCS.getMessages.returns([
+			{
+				code: 'foo-code',
+				element: {
+					parentNode: null,
+					isEqualNode: function() {return false;},
```
GlynnPhillips: I am not keen on this approach but its the cleanest approach I have found so far.

lib/inject.js
```diff
@@ -140,6 +138,25 @@ function injectPa11y(window, options, done) {
 		return false;
 	}
 
+	function hiddenAreasContainsElement(element, hiddenSelectors) {
```
GlynnPhillips: This function fails linting for its complexity, shall I split it up or just add the ignore comments?

lib/inject.js
```diff
@@ -140,6 +138,25 @@ function injectPa11y(window, options, done) {
...
+		var hiddenElements = window.document.querySelectorAll(hiddenSelectors);
+
+		for(var i = 0; i < hiddenElements.length; i++) {
+			if(element.isEqualNode(hiddenElements[i])) {
+				return true;
+			}
+
+			var parent = element.parentNode;
+			while (parent) {
+				if(parent.isEqualNode(hiddenElements[i])) {
```
GlynnPhillips: This fails the linting also for being too deeply nested, again refactor or ignore comments?

lib/inject.js
rowanmanning: Ignore I reckon, I'm not sure if it's possible to simplify. At least not easily. I'd rather have the bug fixed and address complexity later :)

lib/inject.js
rowanmanning: Yep ignore. See above

test/integration/mock/html/hide-elements.html
```diff
@@ -7,6 +7,34 @@
...
+		<p id="para">foo</p>
+		<p id="para">bar</p>
+
+		<form action="">
+			<!-- No label -->
+			<input id="input" name="input" value="" />
+			<!-- No submit -->
+		</form>
+	</div>
 </body>
```
GlynnPhillips: I would love it if people know of some more a11y issues we can chuck in this file to make sure this time this feature is fully functional

GlynnPhillips:
Code style of linting errors have now been fixed but I am not pushing this yet as I want all the above comments to be seen easily for now.

@rowanmanning is helping look into the failing tests which I imagine are due to Phantom 1.9

rowanmanning:
This is brilliant, thanks Glynn 👍 feel free to merge and tag up version `3.8.1`

GlynnPhillips:
I might try and find a couple more a11y errors and pop them in the integration test so reassure myself that we have covered everything correct and then I will push this.

PR #161, opened 2016-05-27
Adds new question for pa11y html codesniffer

nickcall:
Hope to answer the question of why pa11y gives different results to html codesniffer bookmarklet

### DIFF
--- a/README.md
+++ b/README.md
@@ -739,7 +739,7 @@ If parts of the DOM are been loaded after the document is first generated, you m
 If you use Pa11y and HTML CodeSniffer CLI you will find that you get the same results, which will both differ from the bookmarklet, a similar issue was highlighted by [HTML CodeSniffer][sniff-issue].
 
 We are aware of an issue with regard to iframe content, PhantomJS doesn't automatically provide access to the inner content of iframes, and so Pa11y doesn't currently support testing against iframe content from a parent page context.
-Any page that makes use of iframes, e.g. for displaying ads, will show different results on Pa11y than those when running HTML_CodeSniffer in the browser.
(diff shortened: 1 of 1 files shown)


hollsk:
In future I think I'd like to see a troubleshooting.md page for things like this, as it's not really like the others (which are how-tos). Not worth doing it for one question right now, though. 

Maybe create a separate issue to go through all the closed issues looking for things that have cropped up more than once?

rowanmanning:
Also hopefully people will search the issues before raising the same one again, so we should reply to all the issues and link to the relevant docs :)

rowanmanning:
This looks great to me! 👍

README.md
```diff
@@ -731,6 +731,19 @@ pa11y({
...
 
+### Why does Pa11y give different results to HTML CodeSniffer bookmarklet ?
+
+Pa11y uses PhantomJS as a headless web browser to load the DOM content and can only analyse what is provided.
+If parts of the DOM are been loaded after the document is first generated, you may get results that differ from the bookmarklet which runs in the browser and can test against the complete DOM.
+
+If you use Pa11y and HTML CodeSniffer CLI you will find that you get the same results, which will both differ from the bookmarklet, a similar issue was highlighted by [HTML CodeSniffer][sniff-issue].
+
+We are aware of an issue with regard to iframe content, PhantomJS doesn't automatically provide access to the inner content of iframes, and so Pa11y doesn't currently support testing against iframe content from a parent page context.
+Any page that makes use of iframes, e.g. for displaying ads, will show different results on Pa11y than those when running HTML_CodeSniffer in the browser.
```
andrewmee: Bit pedantic, but this should be "may" not "will".  If there are no errors/warnings in the iframe then they'd show the same results either way.

PR #179, opened 2016-07-20
Hover status info added to Common questions

mariabg:


typothecary:
Hi Maria - thank you for contributing. I'm not very technical and haven't got push access to edit.  I have some suggestions for the text.  The guidelines say we should write "informally" and use "active voice" rather than passive.  Also, do we really need to mention CodeSniffer? (It might change in the future).

You wrote: 
      +### Is Pa11y testing the contrast of the hover state ?
      +Pa11y uses HTML CodeSniffer to test the site, and as it does not check the hover state, Pa11y will no do it either. Instead, the contrast of your hover state colours can be tested by adding them to the page.

 I suggest instead: 

     +### Is Pa11y testing the contrast of the hover state?
     +Pa11y doesn't check the hover state. Instead, you must test the contrast of the hover state for links manually.

rowanmanning:
Looks great 👍
Thanks Maria and Kathleen!

PR #184, opened 2016-07-22
add recommendation

### LINKED ISSUE #123: Test with PhantomJS 1.9 and 2.0
I assumed that pa11y worked with PhantomJS 1.9 and 2.0 but I may be wrong. We should investigate this and clarify the versions of PhantomJS that we support in the README.

As an extra, it may even be possible to test with multiple versions of PhantomJS on Travis. If we can then we should.

hollsk:
Add latest stable recommendation to the docs.

Resolves #123.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Table Of Contents
 Requirements
 ------------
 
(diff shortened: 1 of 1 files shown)


rowanmanning:
Looks good 👍 sorry this one slipped through the cracks!

PR #190, opened 2016-09-07
Install phantomjs as a dependency

JakeChampion:
I thought it might be a good idea to have PhantomJS be installed automatically when installing Pa11y.

Users are still able to override the binary path used for phantomJS if they want to supply their own PhantomJS binary.

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -7,6 +7,7 @@ var lowercase = require('lower-case');
 var pkg = require('../package.json');
 var truffler = require('truffler');
 var trufflerPkg = require('truffler/package.json');
+var phantomjsPath = require('phantomjs-prebuilt').path;
 
 module.exports = pa11y;
 module.exports.defaults = {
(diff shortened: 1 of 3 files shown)


petetnt:
👍 This might remove the additional Windows "if you run into..." steps altogether too https://github.com/pa11y/pa11y#windows

joseluisbolos:
@rowanmanning Any updates about this issue?

lib/pa11y.js
```diff
@@ -31,7 +32,8 @@ module.exports.defaults = {
 		parameters: {
 			'ignore-ssl-errors': 'true',
 			'ssl-protocol': 'tlsv1'
-		}
+		},
+		path: phantomjsPath
```
JakeChampion: If this doesn't make it in, would Pa11y be willing to allow the path to be configurable via the API? It already is configurable via the CLI.

rowanmanning:
Thanks @JakeChampion, I've reopened as #207 where I can make a few small tweaks and do some testing.

Issue #181, opened 2016-07-22
Check the documentation for typos


rowanmanning:
I've just noticed that in the README I've spelled "seperated" instead of "separated". There are probably more of these, and I don't want to open a PR with a single letter change 😂

rowanmanning:
I think we can probably close this as @bradarv90 did a great job of catching issues 🙂

PR #194, opened 2016-10-10
#181 | Fixed spelling errors in markdown files

bradarv90:
Fix for #181


hollsk:
Terrifying number of spelling errors. Thanks so much for picking them up, @bradarv90! 

Agree with @typothecary that "analyse" isn't an error _per se_, just British English. We do have American English elsewhere in the docs and I've deliberately left every instance of it alone because it's not wrong, just different. I'd lean towards leaving regional variations in place.

bradarv90:
That was totally my fault. I changed that back to British English in my most recent commit.

PR #211, opened 2016-11-22
Add a proposal and working code for "actions", a new automation feature

rowanmanning:
This is a proposal and working code for "actions", that should hopefully replace the use of `beforeScript` or provide a safer less error-prone alternative. I've added the ability to use predefined (by us) scripts to perform actions on the page before Pa11y runs tests:

```js
pa11y({
    actions: [
        'set field #username to exampleUser',
        'set field #password to password1234',
        'click element #submit',
        'wait for path to be /myaccount'
    ]
});
```

I've added the following actions as a starting point. I'd like to get this in (tagged as a beta feature) then gather wider feedback from the community. I don't think we can predict which actions users need, but I've added enough to perform commonly requested tasks like logging in or clicking an element:

### Click Element

This allows you to click an element by passing in a CSS selector. This action takes the form `click element <selector>`. E.g.

```js
pa11y({
    actions: [
        'click element #tab-1'
    ]
});
```

### Set Field Value

This allows you to set the value of a text-based input or select box by passing in a CSS selector and value. This action takes the form `set field <selector> to <value>`. E.g.

```js
pa11y({
    actions: [
        'set field #fullname to John Doe'
    ]
});
```

### Check/Uncheck Field

This allows you to check or uncheck checkbox and radio inputs by passing in a CSS selector. This action takes the form `check field <selector>` or `uncheck field <selector>`. E.g.

```js
pa11y({
    actions: [
        'check field #terms-and-conditions',
        'uncheck field #subscribe-to-marketing'
    ]
});
```

### Wait For Fragment/Path/URL

This allows you to pause the test until a condition is met, and the page has either a given fragment, path, or URL. This will wait until Pa11y times out so it should be used after another action that would trigger the change in state. This action takes one of the forms:

  - `wait for fragment to be <fragment>` (including the preceeding `#`)
  - `wait for path to be <path>` (including the preceeding `/`)
  - `wait for url to be <url>`

E.g.

```js
pa11y({
    actions: [
        'wait for path to be /login'
    ]
});
```

---

This should all be easier to add into the dashboard/webservice without having to think about the security implications of users being able to run arbitrary JS. Also it lowers the barrier to entry for running code before Pa11y – you don't need to understand JS in order to add actions to your config file.

What do people think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,10 +593,13 @@ Actions are additional interactions that you can make Pa11y perform before the t
 ```js
 pa11y({
     actions: [
-        'set field #username to exampleUser',
-        'set field #password to password1234',
-        'click element #submit',
-        'wait for path to be /myaccount'
+        'click element #tab-1',
+        'set field #fullname to John Doe',
+        'check field #terms-and-conditions',
+        'uncheck field #subscribe-to-marketing',
+        'wait for fragment to be #page-2',
+        'wait for path to be /login',
+        'wait for url to be https://example.com/'
     ]
 });
 ```
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ Table Of Contents
 - [Command-Line Interface](#command-line-interface)
 - [JavaScript Interface](#javascript-interface)
 - [Configuration](#configuration)
+- [Actions](#actions)
 - [Examples](#examples)
 - [Common Questions](#common-questions)
 - [Contributing](#contributing)
@@ -325,6 +326,25 @@ test.run('http://www.nature.com/', {
 
 Below is a reference of all the options that are available:
 
+### `actions` (array) **BETA**
+
+Actions to be run before Pa11y tests the page. There are quite a few different actions available in Pa11y, the [Actions documentation](#actions) outlines each of them.
+
+**Note:** actions are currently in a beta state and the API may change while we gather feedback.
+
+```js
+pa11y({
+    actions: [
+        'set field #username to exampleUser',
+        'set field #password to password1234',
+        'click element #submit',
+        'wait for path to be /myaccount'
+    ]
+});
+```
+
+Defaults to an empty array.
+
 ### `allowedStandards` (array)
 
 The accessibility standards that are allowed to be used. This can be modified to allow for custom HTML CodeSniffer standards.
@@ -339,7 +359,9 @@ Defaults to `Section508`, `WCAG2A`, `WCAG2AA`, and `WCAG2AAA`.
 
 ### `beforeScript` (function)
 
-A function to be run before Pa11y tests the page. The function accepts three parameters;
+**Note:** unless you're doing something particularly complicated, it's much easier and less error prone to use [actions](#actions) rather than `beforeScript`. If you specify both, the `beforeScript` will be dropped with a warning.
+
+A function to be run before Pa11y tests the page. The function accepts three parameters:
 
 - `page` is the phantomjs page object, [documentation for the phantom bridge can be found here][phantom-node-options]
 - `options` is the finished options object used to configure pa11y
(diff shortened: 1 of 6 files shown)


GlynnPhillips:
I like this idea and it's more inline with browser automation libraries such as [Nightwatch](http://nightwatchjs.org/).

Have you already considered what other automation libraries there are for phantom that we might be able to use so we have less to maintain? Only one I know of off the top of my head is [Casper](casperjs.org) and I don't think that would be a good fit.

hollsk:
If we were to use a Phantom automation library, it'd couple us more tightly to Phantom, which isn't ideal as long-term we'd like to give users more choice of headless browser. 

I do really like the `actions` idea, it'd make things a lot easier for people.

GlynnPhillips:
@hollsk Good point and agree I do much prefer `actions` to `beforeScript`

hollsk:
Linking the related ticket in [pa11y/webservice](https://github.com/pa11y/webservice/issues/42) that kicked off the discussion

una:
I love this idea -- definitely an easier-to-use beforeScript! 

I'd like to keep in mind that binding on values other than an ID are possible. I.e. 

`<div data-js="login-form">`  -->    `'click element [data-js="login-form"]'`
`<input class="login f_name">` --> `'click element .login.f_name'`

Are there any selectors in which this wouldn't work and we might still want to use js-syntax for variables?

rowanmanning:
Hey Una, I've tried to ensure that this works with any selector so all of the ones you mention should be fine. Part of the reason I've labeled this as a beta feature to start with is so that we can gather feedback from real users – I can't think of every case.

Actually I have some code I was using during testing that has examples of longer selectors. This opens GitHub, clicks the login button, enters some incorrect auth details and then tests the session error page:

```js
'use strict';

var pa11y = require('.');

var test = pa11y({
	log: {
		debug: console.log.bind(console),
		error: console.error.bind(console),
		info: console.log.bind(console)
	}
});

test.run('https://github.com/', {
	actions: [
		'click a.btn.site-header-actions-btn.mr-1',
		'wait for path to be /login',
		'set field #login_field to exampleUser',
		'set field #password to password1234',
		'click [name=commit]',
		'wait for path to be /session'
	]
}, function(error) {
	if (error) {
		return console.error(error.message);
	}
	console.log('Not logging results');
});
```

hollsk:
I've only got one comment and it's on the docs, of course :p 

Everything looks very well-thought out.

README.md
```diff
@@ -563,6 +585,80 @@ pa11y({
...
 
 
+Actions
+-------
+
+Actions are additional interactions that you can make Pa11y perform before the tests are run. They allow you to do things like click on a button, enter a value in a form, wait for a redirect, or wait for the URL fragment to change:
+
+```js
+pa11y({
+    actions: [
```
rowanmanning: Good idea, I've made this change

GlynnPhillips:
I think it looks great, nice work @rowanmanning

ray-kim-12:
@rowanmanning tried your snippet, but not working and not logging any results.
```
  pa11y(LOGIN_PAGE, {
    actions: [
      'wait for element #user_username to be visible',
      'click element #user_username',
      'set field #user_username to USERNAME',
      'click element #user_password',
      'set field #user_password to PASSWORD',
      'click element #signIn',
      'wait for url to be http://localhost:9000/'
    ]
  }, function(err) {
    if (err) {
      return console.error(err.message);
    }
    console.log('not logging results')
  })
  console.log('page.url() should be localhost:9000', page.url())
```

the `actions` feature seems very flakey. 
I can log in via puppeteer but when I try to access a page behind auth with pa11y like  `pa11y(page_behind_auth)` it's not able to get it.

PR #213, opened 2016-11-24
Add snyk policy file and snyk test integration

joseluisbolos:


package.json
```diff
@@ -54,6 +54,6 @@
 		"pa11y": "./bin/pa11y"
 	},
 	"scripts": {
-		"test": "make ci"
+		"test": "snyk test && make ci"
```
rowanmanning: How does the `snyk` command get installed? Maybe it should be in dev dependencies in `package.json`? Otherwise people have to remember to install it themselves.

Also `npm test` is never run in CI – so it shouldn't go here. It should be in [`.travis.yml`](https://github.com/pa11y/pa11y/blob/master/.travis.yml#L27) I think :)

joseluisbolos: Actually, all of this has been added by the snyk CLI automatically. Seems that it's not smart enough to realise that we're running `make test` in Travis instead of `npm test` (which probably we should be using anyway - it seems bad practice having to specify the same command in two different places).

Let me study this it a bit further and we can discuss it in Slack later.

PR #216, opened 2016-11-25
Display: Document title in debug mode

andygout:
cc @lc512k @rowanmanning 

With `--debug` flag turned on, running `pa11y` will add to the debug logs the document title of the page tested.

This will allow user to ensure that passing tests are not false positives called from error pages which may not cause any pa11y errors/warnings while the actual page might. Naturally, this relies on the website being tested having appropriate document titles when a page errors (but given the variety of forms these may take it still requires human judgement to deduce whether test has been conducted on the intended page).

### Page found
page-found

### Page not found
page-not-found

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 
 # Changelog
 
+## 4.1.1 (2016-11-24)
+
+  * Update dependencies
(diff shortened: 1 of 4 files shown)


lc512k:
Great work, @andygout! This will really help.

PR #222, opened 2016-12-06
Return: Entire result object (inc. document title)

andygout:
cc @rowanmanning @GlynnPhillips @lc512k

The suggested format change of returned might be problematic if other elsewhere requires it in current form (I have only accommodated the change in `pa11y/ci` in this PR: https://github.com/pa11y/ci/pull/9).

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -64,7 +64,7 @@ function testPage(browser, page, options, done) {
 			return done(new Error(result.error));
 		}
 		options.log.debug('Document title: "' + result.documentTitle + '"');
-		done(null, result);
+		done(null, result.messages, result.documentTitle);
 	});
 
 	async.waterfall([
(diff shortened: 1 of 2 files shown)


rowanmanning:
Thanks Andy :) Yep you're right, this would be a breaking change – quite a few Pa11y users use the JavaScript API directly and their code would break. I think in order to avoid that, we could change the done call:

```js
done(null, result.messages, result);
```

That way existing users will continue to get just the messages, and new users can grab the full results from the third callback param (including us in Pa11y CI).

andygout:
Adding a third parameter to the [`done()` callback](https://github.com/pa11y/pa11y/blob/master/lib/pa11y.js#L67) does not get recognised when [received](https://github.com/pa11y/ci/blob/master/lib/pa11y-ci.js#L74); it appears as `undefined`.

Think I might be missing something obvious...

rowanmanning:
Oh, yeah, because of `async`. Agh, that's annoying. Might be worth pairing on this at some point this week? I know there will be a way around it but I have no idea how right now 😄

andygout:
Will implement in new way as discussed.

PR #220, opened 2016-12-05
Conversion of JSHint and JSCS to ESlint #218

pamo:
Used [Polyjuice](https://www.npmjs.com/package/polyjuice) to aid in conversion.
Updated the exception comments to match eslint one-line disable/enable.
Removed the code-style step in Makefile to consolidate into one lint step.

Happy to make additional updates to close off #218.

Issue #218, opened 2016-11-26
Replace JSCS


joseluisbolos:
JSCS has merged with ESLint and it seems like it's not under development any more. We should consider replacing it.

rowanmanning:
Yeah I've been meaning to for a while, it just seems like a big job to make sure the configs match 😁. We can actually replace both JSCS _and_ JSHint with ESLint – it does the job of both.

PR #223, opened 2016-12-06
Cleans up remaining jshint and jscs disable flags

pamo:
Realized I missed a few unnecessary disable flags in the previous PR (#220) I opened regarding #218.


rowanmanning:
Thanks @pamo :) good spot

PR #224, opened 2016-12-14
Add: Verify page flag

andygout:
cc @GlynnPhillips @rowanmanning @lc512k

Allows user to add a string of HTML that they expect to be included in the HTML of the page being tested to verify that the intended page is being tested (i.e. not an Application Error page) and/or certain elements are present on the page.

Next step is to add similar flag in `pa11y-ci` that can pass the option through into `pa11y`: https://github.com/pa11y/ci/pull/11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -777,7 +777,7 @@ If you're opening issues related to these, please mention the version that the i
 License
 -------
 
-Pa11y is licensed under the [Lesser General Public License (LGPL-3.0)][info-license].
+Pa11y is licensed under the [Lesser General Public License (LGPL-3.0)][info-license].<br>
 Copyright &copy; 2016, Springer Nature
 
 
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Usage: pa11y [options] <url>
     -p, --port <port>             the port to run PhantomJS on
     -t, --timeout <ms>            the timeout in milliseconds
     -w, --wait <ms>               the time to wait before running tests in milliseconds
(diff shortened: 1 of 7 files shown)


lib/inject.js
```diff
@@ -3,6 +3,13 @@
 /* eslint-disable max-statements */
 function injectPa11y(window, options, done) {
 
+	if (options.verifyPage) {
+		var windowHtml = window.document.documentElement.outerHTML;
+		if (!windowHtml.match(new RegExp(options.verifyPage))) {
+			return reportError('Page not verified');
```
rowanmanning: I'd maybe make this more descriptive. Something like `The page can not be verified, it was expected to match the following regular expression: <regexp>`?

README.md
```diff
@@ -764,7 +777,7 @@ If you're opening issues related to these, please mention the version that the i
 License
 -------
 
-Pa11y is licensed under the [Lesser General Public License (LGPL-3.0)][info-license].  
+Pa11y is licensed under the [Lesser General Public License (LGPL-3.0)][info-license].
```
rowanmanning: This change will break the formatting of the README. A double space at the end of a line is a `<br/>`. I think as this keeps happening we should just add in an actual `<br/>` 😄 could you do that?

andygout: 👍  Have gone with `<br>` (sans forward slash) based on [Stack Overflow](http://stackoverflow.com/questions/1946426/html-5-is-it-br-br-or-br), unless you have a preference for how you've given i?

rowanmanning:
Thanks for this! Released `4.3.0`

PR #230, opened 2017-01-24
Expose action validation (for use in webservice)

rowanmanning:
Ideally we want to be able to check whether an action is valid before using it. This is especially useful in Pa11y webservice which would otherwise fail pretty much silently when an invalid action is encountered.

This adds the following method to pa11y:

```js
pa11y.validateAction('click element #submit');  // true
pa11y.validateAction('open the pod bay doors'); // false
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,17 @@ var htmlReporter = require('pa11y/reporter/html');
 var html = htmlReporter.process(results, url);
 ```
 
+### Validating Actions
+
+Pa11y exposes a function which allows you to validate [action](#actions) strings before attempting to use them.
+
+This function accepts an action string and returns a boolean indicating whether the action is valid or not:
+
+```js
+pa11y.validateAction('click element #submit');  // true
+pa11y.validateAction('open the pod bay doors'); // false
+```
+
 
 Configuration
 -------------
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -295,6 +295,17 @@ var htmlReporter = require('pa11y/reporter/html');
 var html = htmlReporter.process(results, url);
 ```
 
+### Validating Actions
+
+Pa11y exposes a function which allows you to validate [action](#actions) strings before attempting to use them.
+
+This function accepts an action string and returns a boolean indicating whether the action is valid or not:
```
hollsk: I had to go back and check to try and puzzle out what you meant by "valid" 😁 

Maybe be explicit that "valid" means "specified in the actions API"

README.md
```diff
@@ -295,6 +295,17 @@ var htmlReporter = require('pa11y/reporter/html');
...
+
+```js
+pa11y.validateAction('click element #submit');  // true
+pa11y.validateAction('open the pod bay doors'); // false
```
hollsk: 😄

rowanmanning:
How about now, @hollsk? Not sure if it makes sense, is it better?

hollsk:
Yeah, looks great 👍

Issue #228, opened 2017-01-20
Decide on new actions to implement


rowanmanning:
Now that [actions](https://github.com/pa11y/pa11y#actions) have launched, we should think about new ones we could add that will make life easier for single-page applications and testing modifications to the page made by JavaScript. So far the following are what we have:

  - [x] `click element <selector>`
  - [x] `set field <selector> to <value>`
  - [x] `check field <selector>`
  - [x] `uncheck field <selector>`
  - [x] `wait for fragment to be <fragment>`
  - [x] `wait for path to be <path>`
  - [x] `wait for url to be <url>`

I can think of a few ideas for new ones off the top of my head:

  - [x] `wait for element <selector> to be added`
  - [x] `wait for element <selector> to be removed`
  - [x] `wait for element <selector> to be visible`
  - [x] `wait for element <selector> to be hidden`
  - [x] `wait for element <selector> to emit <event>` (only in Pa11y 5.0+)

New ideas from the issue comments:

  - [x] `screencapture example.png` (only in Pa11y 5.0+)

hollsk:
We can get some ideas from previous issues - e.g. for issue #199 this would be needed:

`wait for element <selector> innerHTML to be <value>`

rojlarge:
Hi all-- first off, love the project! Automating accessibility testing with pa11y has been hugely useful for my team and was pretty easy to get set up.

One situation that we've run into frequently though is the need for all of our tests to be able to log in through a single sign-on portal, which isn't very well supported by the current "wait for..." actions. It would be awesome if we could specify an action like `wait for path to not be /sso/login` or something similar so that all of our tests could use the same set of actions without having to hardcode the url or path under test.

I was thinking something along the lines of this: https://github.com/pa11y/pa11y/compare/master...rojlarge:wait-for-action-negation. If that sounds like something you're interested in, just let me know and I'd be happy to submit a PR for it.

rowanmanning:
Hi @rojlarge, that's great! Please do submit a PR, the only thing I can see that's missing is an update to the documentation

iandevlin:
An enhancement to the `actions` would be ideal for my use case right now, as I would like to send a `click` action and then wait for some JS loaded content to load before running pa11y. So this enhancement gets a +1 from me!

iandevlin:
Today I spent some time attempting to implement a `wait-for-load` action that takes the form `wait for .foo .bar to load` before running pa11y.
[...]

PR #231, opened 2017-01-26
Add negated version of "wait-for-url" action

rojlarge:
This updates the `wait-for-url` action to allow test execution to be paused until the page does **not** match a given fragment, path, or URL if "to not be" is included in the action expression as discussed in [#228](https://github.com/pa11y/pa11y/issues/228#issuecomment-275021330).

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,7 +609,7 @@ pa11y({
         'check field #terms-and-conditions',
         'uncheck field #subscribe-to-marketing',
         'wait for fragment to be #page-2',
-        'wait for path to be /login',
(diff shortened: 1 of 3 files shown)


rojlarge:
The integration test that failed here doesn't seem to be related to these changes, and I can't reproduce the failure locally when I run the ITs against Node 4. Can you re-trigger the build and see if it fails again?

rowanmanning:
Thanks @rojlarge, this is great! Merging now and will cut a release ASAP

PR #243, opened 2017-03-16
rejig guides and add actions recommendation

hollsk:
I've moved the common questions section out of the readme and into the troubleshooting doc, and made some changes to the AngularJS section to recommend actions as a way of overcoming the ngViewLoaded problem. I've also done some minor reformatting of the common problems section so it's consistent with the rest of the doc.


README.md
```diff
@@ -725,130 +725,10 @@ Step through some actions before Pa11y runs. This example logs into a fictional
 Inject a script before Pa11y runs. This example logs into a fictional site then waits until the account page has loaded before running Pa11y. [See the example](example/before-script/index.js).
 
 
-Common Questions
-----------------
+Common Questions and Troubleshooting
```
rowanmanning: As this title has changed, the ID of the heading element will have too. This needs to be updated in the table of contents

hollsk: Good catch! Updated :D

PR #250, opened 2017-03-20
Use ESLint with the pa11y lint config

rowanmanning:


lib/inject.js
```diff
@@ -134,7 +133,7 @@ function injectPa11y(window, options, done) {
 			return hiddenAreasContainsElement(element, selector);
 		});
 
-		return elementsWithinHiddenSelectors.length ? false : true;
+		return !elementsWithinHiddenSelectors.length;
```
andrewmee: Is this necessary for the linter?  It doesn't read anywhere near as clearly IMO

rowanmanning: Happy to change this rule to a warning rather than an error. I don't want to get rid of the rule entirely because [it picks up on more useful stuff too](http://eslint.org/docs/rules/no-unneeded-ternary).

So either a warning, or in this specific case we add a comment to disable the rule? Maybe I'm leaning more towards a comment for this specific case...

andrewmee: A comment to opt-out here seems sensible to me

PR #253, opened 2017-03-28
[Feature] add option to cherry-pick WCAG rules

Lily2point0:
New feature that allows users to pass specific guidelines from a higher WCAG 2.0 standard

Code:
- Adds new option flag `-A` or `--add-rule`
- Checks if rule is valid (against WCAG 2.0 AAA list in HTMLCS)
- Adds rule window object for current standard
- Throws error if rule is invalid
- Ignores flag if standard is Section508

Tests:
- Add integration test, `cli-rules`
- Update unit tests `inject.js`


README.md
```diff
@@ -111,6 +111,7 @@ Usage: pa11y [options] <url>
     -H, --htmlcs <url>             the URL or path to source HTML_CodeSniffer from
     -e, --phantomjs <path>         the path to the phantomjs executable
     -S, --screen-capture <path>    a path to save a screen capture of the page to
+    -A, --add-rule <rule>          WCAG 2.0 rules from a different standard to include, a repeatable value or separated by semi-colons
```
rowanmanning: Could we also add documentation for the JavaScript API in https://github.com/pa11y/pa11y#configuration ?

Lily2point0: Sure, will get on that right away. 
Re: CI, I have a random fail that shows only some times, and not always in `lib/pa11y`; something to do with the `afterEach` method?

PR #235, opened 2017-02-15
Wait for element to be action

iandevlin:
As discussed, this is a work in progress PR for adding a `wait-for` action.

PR #262, opened 2017-05-12
added support for the wait-for-url 'host' field

ondras:
This PR enriches the "wait-for-url" functionality by adding an option to compare just the "host" value. This is useful for scenarios when your setup includes cross-domain redirects, but you do not know those full URL values.


rowanmanning:
This is great, thanks @ondras! I'll try to get it merged and released soon alongside #235

PR #304, opened 2017-08-23
Switch from PhantomJS to Headless Chrome

rowanmanning:
This PR replaces PhantomJS with Headless Chrome. This will hopefully go into the 5.x branch and be released later with the rest of that stuff. I've opened up several issues in the [5.0 milestone](https://github.com/pa11y/pa11y/milestone/7) to address the `TODO` items that appear in this PR.

If you want to do a "first-pass" review, I'd recommend looking at the README and MIGRATION file changes mostly.

Sorry for the huge amount of code, I think after this I may release a beta so that people can try it out. I think it'll be easier to just [look at the code on the branch](https://github.com/pa11y/pa11y/tree/headless-chrome) rather than review the changes as most of the JS has had to be completely rewritten.

### Why Node.js 8+ only?
We wanted to switch to using the most modern development methods we can. We also wanted to use [Puppeteer](https://github.com/GoogleChrome/puppeteer), which requires Node.js 7.10+. We'll be keeping Pa11y 4.x maintained for quite some time (possibly as far as April 2019 when Node.js 6.x maintenance ends) and will update everyone when we decide on an end-of-support date.

---

(I've just noticed this is #304 😁 definitely not a status code that describes this work)


andrewmee:
All the documentation looks good 👍  

Will try to get some time to dive in to the code for review soon (if someone doesn't get to that first)

hollsk:
I've set myself a reminder to have a look tomorrow, but the more the merrier 😁

lib/action.js
```diff
@@ -1,72 +1,57 @@
-'use strict';
```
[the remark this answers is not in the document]
rowanmanning: Oh well spotted, I would have forgotten! Thanks for the review, Lily :)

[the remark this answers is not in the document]
TheJaredWilcurt: Why not just have everything pass a linter before allowing it to be merged?

* https://eslint.org/docs/rules/strict

MIGRATION.md
```diff
@@ -8,17 +8,67 @@ Pa11y's API changes between major versions. This is a guide to help you make the
...
+
+**TODO this migration guide is a work in progress. Please update as you work on Pa11y 5.x**
+
+### PhantomJS to Headless Chrome
+
+Pa11y 5.0 switches from PhantomJS to [Headless Chrome](https://developers.google.com/web/updates/2017/04/headless-chrome). This allows us to use more modern JavaScript APIs and make Pa11y testing more stable.
+
+### Node.js Support
+
+Pa11y 5.0 now only supports Node.js v8.0.0 and higher, you'll need to upgrade to be able to use the latest versions of Pa11y.
```
hollsk: I'd change "now only supports" to "only supports" to remove any kind of timeline ambiguity (like, "when is now?" philosoraptor.jpg)

MIGRATION.md
```diff
@@ -8,17 +8,67 @@ Pa11y's API changes between major versions. This is a guide to help you make the
...
+
+### Command-Line Interface
+
+The command-line interface in 5.0 is similar to 4.0, but there are a few key changes.
+
+  - The `--verify-page` flag has been removed, as page verification can be achieved with screen-shots or observing the debug output
+  - The `--htmlcs` flag has been removed, you can no longer configure the version of HTML CodeSniffer that Pa11y uses (TODO reconsider this)
```
hollsk: In two minds. On the one hand it'd be useful for people who want to upgrade/downgrade htmlcs (could this have saved us during our second-to-last release? 🤔 ). 

On the other hand maybe it'd be better to come up with an API of sorts so people can plug in $ENGINE and have it produce standard output.

MIGRATION.md
```diff
@@ -8,17 +8,67 @@ Pa11y's API changes between major versions. This is a guide to help you make the
...
+Pa11y is now completely [`Promise`-based](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise), and uses [`async`/`await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) internally. This means the API has changed significantly.
+
+You no longer need to create a separate test function, now Pa11y is just one function: `pa11y(url, options)`. See the [README](README.md) for more information.
+
+### Configuration
+
+Configuration options have had an update between 4.0 and 5.0:
+
+  - The `allowedStandards` option has been removed. This can still be set by manually adding entries to `pa11y.allowedStandards`
+  - The `beforeScript` option has been removed in favour of using actions
```
hollsk: Maybe add a link to #228 with a note saying 'if something's broke, tell us there'? 😁

MIGRATION.md
```diff
@@ -8,17 +8,67 @@ Pa11y's API changes between major versions. This is a guide to help you make the
...
+  - The `page` option has been removed, as this was specific to PhantomJS. You can set the following options to achieve the same results:
+    - `page.headers` can now be set with a new `headers` option
+    - `page.settings.userAgent` can now be set with a new `userAgent` option
+    - `page.viewport` can now be set with a new `viewport` option
+  - The `phantom` option has been removed, as this was specific to PhantomJS
+  - The `verifyPage` option has been removed, as page verification can be achieved with screen-shots or by inspecting the `documentTitle` property of the results
+
+### Miscellaneous Changes
+
+  - A full scheme is now required in Pa11y URLs, you can no longer run `pa11y example.com`, you need to use `pa11y http://example.com/` (TODO reconsider this)
```
hollsk: I don't have a problem with this change. I always use the scheme anyway, IIRC - maybe other reviewers have an opinion?

TROUBLESHOOTING.md
```diff
@@ -1,82 +1,64 @@
...
 
 To debug this, you can use the `--screen-capture` flag or [`screenCapture` option](https://github.com/pa11y/pa11y#screencapture-string) and compare the output for the differing test runs.
 
 ### Pa11y ignores the contents of iframes
-Pa11y doesn't currently support testing against iframe content from a parent page context. Any page that makes use of iframes, e.g. for displaying ads, may show different error, warning, and notice counts in Pa11y compared to the HTML_CodeSniffer browser bookmarklet.
-
-If you do need to test the contents of an iframe, run Pa11y against the iframe source URL directly.
 
-### PhantomJS fails to run and returns exit code 127
+TODO verify that this is still an issue with Headless Chrome
```
hollsk: We won't know if headless chrome has similar problems until it's being used out in the wild - even if it does have similar problems, they won't really look like this one. 

Maybe we should leave this section alone so that 4.x users can still refer to it, until 4.x is retired

TROUBLESHOOTING.md
```diff
@@ -86,107 +68,37 @@ pa11y({
...
-        }, function() {
-
-            waitUntil(function() {
-                return dynamicContent.getAttribute('aria-hidden') === 'false';
-            }, 20, next);
-        });
-    }
-});
-```
+We recommend using [actions][actions].
```
hollsk: This would be a good place to plug #228 again 😁

lib/action.js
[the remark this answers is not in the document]
rowanmanning: @TheJaredWilcurt: we do use a linter across the projects, but the contribution guidelines outline some basic principals that contributors can stick to in order to make their lives easier later.

@Lily2point0: I've decided I'm just gonna add the `'use strict';` back in 😁

MIGRATION.md
rowanmanning: Yeah I just feel like it might be an unnecessary removal, and cause loads of questions. It's only a couple of lines of code to add it back in and maintain backwards-compatibility with more of the CLI.

MIGRATION.md
rowanmanning: Yeah I think _maybe_ it's easier to remove this and readdress if somebody needs to switch out Phantom. Technically they still can but just with their own fork/build of Pa11y (if the need is dire and we don't have engine switching figured out.

rowanmanning:
OK @hollsk: I've updated the docs based on all your changes/comments.

@Lily2point0: I've added `use strict` back in, I'm not 100% sure why I removed it anyway!

Thanks so much all of you for the reviews, your feedback is really helpful :)

PR #309, opened 2017-08-28
bubbles an event on input value change

### LINKED ISSUE #306: wait for element not working for angular
Once I noticed wait for element....added to 4.12.1 version I immediately tried and it looks good for non angular pages but not working for single page application. I tried wait for element #id to be added / wait for element #id to be visible. 
```
'actions': [
   		        'set field #locationInput to Irving, TX',
   		        'click element #search',
   		        'wait for element #searchResultContainer to be visible'
   		  	]
```
I am always getting '`Failed action: element #searchResultContainer to be visible'` log error. 

**Here is angular code in same page** 
```
..input .....
...search button
<!-- Result Container -->
<div id="searchResultContainer " ng-show="myController.results > 0">
...other logic
</div>
```
I know searchResultContainer  always exist in dom. I am wondering is something we can do element displayed or like that 

I tried from ng-if condition and `wait for element #searchResultContainer  to be added`, even though I am not getting success 
```
<div id="searchResultContainer " ng-if="myController.results > 0">
.......
</div>
```

This new feature working for non angular pages perfectly. I am not sure If I am missing anything. Would you please check this one, as this feature gonna be very useful for lot of users. Thanks

hollsk:
Hopefully resolves #306 - AngularJS doesn't watch for changes to text fields without additional controller logic, so if we naively set a value for an input field in Pa11y, nothing happens in the application being tested. 

This PR just adds an event to go along with the value change. 

Edit: oh i forgot about the tests 😁


hollsk:
Right. I've broken it and I don't know how to fix it 😭 

* `new Event()` gives me `ReferenceError: Event is not defined`
* `document.createEvent('CustomEvent')` gives me `TypeError: document.createEvent is not a function`
* `new CustomEvent('inputEvent')` gives me `ReferenceError: CustomEvent is not defined`

A little halp, please, someone 😆

rowanmanning:
First, cool, well done for finding! I need to think how we can write some integration tests around this. I think your issue _should_ just be broken tests rather than the code itself, which is good.

In the unit tests, we have to completely mock the DOM stuff that we use, you can see the tests for the bit of code you're updating here:

https://github.com/pa11y/pa11y/blob/master/test/unit/lib/action.js#L434-L457

The `element` variable in the `beforeEach` is our mock DOM element, so you should be able to give that object a `dispatchEvent` method which is a sinon stub.

Then you'll need to mock the global `Event` function, I'd do this in the same way as we set up and tear down `document` in the code above.

Once you've got those things mocked then you could add in a new test to verify that the element dispatches a new event.

Give me a shout if you'd like a hand with this (maybe on Slack), or if I haven't explained clearly enough :)

hollsk:
Ooooh, that helps a _lot_, yes! It works absolutely fine in Phantom, and is only breaking in the tests and I couldn't understand at all what was happening. Knowing that Sinon needs a mock DOM is the missing info, thanks so much 😄 

Will slack you if I get stuck 😁

andrewmee:
Was just discussing with Hollie as we paired on this; I think actually this problem would potentially apply not just to Angular applications, but any app that makes changes based on the state of form fields within in.

Because we're currently just directly updating/manipulating field values those actions would be happening silently.  To help make these actions a bit more realistic and close this gap we'll need to make sure they are as close to a real interaction as possible (so adding any relevant events).  This will need to be done for checkbox/radio button selecting too

rowanmanning:
@andrewmee, @hollsk: ha! I think we've come to this conclusion separately, I was just about to ping you about it. I'm reimplementing actions fully in 5.x, and while integration testing I realised that events weren't being fired and so my code wasn't working.

You're right: text inputs, selects, and textareas all need to fire the `input` event. Checkboxes and radios need to fire the `change` event. The new unit tests exposed that this has been wrong pretty much since we built it :/

hollsk:
thank you for your help both ❤️

mayankagarwal8188:
Hi,

I am using pa11y beta version - 5.0.0-beta.8

I have added an action :
set field input to "value"

But the problem I am facing is on input I am clicking on a button which gets enabled on input keyup event or when the user types in the values. So I am not able to test this scenario, Since the button is not getting enabled as it is just setting the value to the input field and dispatching the input event from action.

Is there any way we can trigger keyup or any other events when setting a value to input field ?

@rowanmanning

PR #312, opened 2017-09-01
Integration test overhaul

### LINKED ISSUE #296: Overhaul the integration tests
Currently there are a load of exceptions in the coding style and I think we can now move to a promises/async way of running these tests. It's probably also worth reviewing how useful some of them are and patching any holes in our coverage.

rowanmanning:
In this PR I've fully rewritten the unit tests for Pa11y, there's a lot here but I think a general review of one or two of the files should be enough. I've been testing extensively as I go and we now have more coverage of the CLI tool (especially actions) than we did before.

This resolves #296


test/integration/cli/actions-check-field.js
```diff
@@ -0,0 +1,71 @@
...
+					'--config', './mock/config/actions-check-field-on.json',
+					'--reporter', 'json',
+					'--ignore', 'warning;notice' // This is so we only deal with errors
+				]
+			});
+		});
+
+		it('checks the field before running tests', () => {
+			assert.isArray(pa11yResponse.json);
+			assert.lengthEquals(pa11yResponse.json, 0);
```
andrewmee: These very indirect looking success cases are rather confusing on first glance. I wonder if a comment in the file might help make it all a bit less arcane.  E.g. something like "The test file has an a11y error in the markup. When this action is performed the DOM is manupulated by JavaScript to remove the offending element, hence no a11y errors is proof of this successful action"

It might be a bit repetitive, but could alleviate some serious head-scratching when looking at any of these file in isolation.

test/integration/mock/html/root-element.html
```diff
@@ -1,16 +1,22 @@
 <!DOCTYPE html>
-<html lang="en">
+<!-- Missing lang attribute generates an error -->
+<html>
 <head>
 
 	<meta charset="utf-8"/>
-	<title>Root Element</title>
+
+	<!-- Page title generates a notice -->
```
andrewmee: Love these comments ❤️

PR #318, opened 2017-09-07
Ignore warnings and notices by default

### LINKED ISSUE #170: Hide warnings and notices by default
I'm not a fan of the warnings and notices that Pa11y outputs. They're not really actionable, and they add quite a bit of noise to the reports. I'm wondering whether we should exclude them.

There are two options I can see:
1. Exclude them by default, but allow them to be enabled with something like `--include warnings`
2. Exclude them always, so you will never see them

Option 2 has the advantage of simplifying the code and allowing us to remove things like the `--level` flag which is all about tweaking what Pa11y should do when it encounters warnings.

Either way this would be a breaking change – something to consider for Pa11y 5.0.

rowanmanning:
We now ignore warnings and notices by default, they get added into the
ignore array once all the options have been defaulted.

I've added a way to continue including them if people want that, this is
just a change in the default behaviour.

Resolves #170

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,8 @@ Usage: pa11y [options] <url>
     -l, --level <level>            the level of issue to fail on (exit with code 2): error, warning, notice
     -T, --threshold <number>       permit this number of errors, warnings, or notices, otherwise fail with exit code 2
     -i, --ignore <ignore>          types and codes of issues to ignore, a repeatable value or separated by semi-colons
-    --include-non-errors           Include warnings and notices in the report
+    --include-notices              Include notices in the report
+    --include-warnings             Include warnings in the report
     -R, --root-element <selector>  a CSS selector used to limit which part of a page is tested
     -E, --hide-elements <hide>     a CSS selector to hide elements from testing, selectors can be comma separated
     -c, --config <path>            a JSON or JavaScript config file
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -171,13 +172,13 @@ For more information on configuring Pa11y, see the [configuration documentation]
 The ignore flag can be used in several different ways. Separated by semi-colons:
 
 ```
```
hollsk: is there a reason for changing this to issue-code-_n_? If I were browsing the docs right now looking for info on this I'd be like "wait, what's issue-code-1? how do i find that out?" as it's not really obviously documented. It's described in the results object below, but you'd need to read down to a totally different section to figure it out

rowanmanning: I mostly wanted to switch away from warning/notice because they're ignored by default and there's no point adding them here. What would you suggest as an alternative? Use real issue codes?

README.md
```diff
@@ -423,19 +409,30 @@ pa11y('http://example.com/', {
...
     ]
 });
 ```
 
 Defaults to an empty array.
 
+### `includeNonErrors` (boolean)
+
+Whether to include non-error results in the Pa11y report. Issues with a type of `warning` or `notice` are not directly actionable and so they are excluded by default. You can include them by using this option:
+
```
hollsk: what do i do if i want warnings but not notices?

rowanmanning: You'd do: ```js
pa11y('http://example.com/', {
    includeNonErrors: true,
    ignore: ['notice']
});
```

Reckon we need to be documented explicitly, or should we wait and see if anyone asks how to do it?

hollsk: always document explicitly 😁

README.md
hollsk: honestly i'd think about removing the ignore flag from the CLI entirely. 

As it is, people would need to switch on non-errors and then also potentially ignore some error code in the same shell command. seems like a bit of a weird and confusing thing to do and even documenting it will be quite confusing :/

hollsk: i am sick though, so maybe it's just that my brain is failing to work properly

rowanmanning: Hmm, I'm not sure how I feel about removing it if we _can_ support it because of how it would break. I wanted to keep the CLI with as few a changes as possible (it's the most used thing) and this one would just exit a Pa11y process immediately if an `--ignore` flag was present.

README.md
rowanmanning: done 👍

README.md
hollsk: I think it's the circularity of it that's bugging me. Default to don't show this, but show this, but don't show that other thing that's part of this. 

What if there was an explicit `--show warning --show notice` (or `--show-error-code-_n_`, i suppose, but I still think that _n_ needs to be explicitly documented at this stage) instead of `--ignore`? Would that cause too many problems with backwards compatibility?

rowanmanning: I guess something like `--include-warnings` and `--include-notices` would be alright, so basically just split the `--show-non-errors` in two?

hollsk: yeah, i think that would be clearer. From a user perspective that would just reverse the 4.x settings which seems easier to understand

rowanmanning: Done 👍

PR #320, opened 2017-09-08
Allow setting the HTTP method and POST data

### LINKED ISSUE #302: Allow POST requests with Headless Chrome
This is for an entry in the troubleshooting guide, which currently has a TODO item.

**Update:** This will require some code, and we'll need to use [Puppeteer's `setRequestInterceptionEnabled` method](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagesetrequestinterceptionenabledvalue).

rowanmanning:
This resolves #302, adding in new options: `method` and `postData`. I've updated the docs and this is probably the best place to look for more info.

A nasty hack has been added to `bin/pa11y.js` to catch uncaught exceptions caused by this issue in Puppeteer: https://github.com/GoogleChrome/puppeteer/issues/627 – once it's fixed then we can update and remove the hack.


bin/pa11y.js
```diff
@@ -9,6 +9,18 @@ const program = require('commander');
 const pa11y = require('../lib/pa11y');
 const semver = require('semver');
```
hollsk: should we create an issue in the pa11y repo for this and link to it here? can't decide 😁 i feel like maybe we'll forget if we can't see it in the issues list

rowanmanning: Good idea 👍  will in a sec if you haven't (haven't checked :D)

PR #327, opened 2017-09-12
remove survey, add 5.0 beta

hollsk:
😱


rowanmanning:
As this is quite long, wondering whether we should just ditch the Sidekick notice from here. It's more applicable to dashboard anyway so not sure whether it's needed any more.

Also there's not been a lot of feedback on it anyway :)

hollsk:
Yeah, I was thinking the same thing!

PR #330, opened 2017-09-14
Add an action to capture a screen-grab

rowanmanning:
This allows a Pa11y user to verify that the page looks correct at different points in the test run.

### DIFF
--- a/lib/action.js
+++ b/lib/action.js
@@ -120,6 +120,20 @@ module.exports.actions = [
 		}
 	},
 
+	// Action to screen capture the page to a file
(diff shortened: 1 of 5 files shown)


lib/action.js
```diff
@@ -120,6 +120,20 @@ module.exports.actions = [
 		}
 	},
 
+	// Action to screen capture the page to a file
```
hollsk: WOW! This is virtually no code at all!

rowanmanning: Headless Chrome was a _great_ idea

PR #339, opened 2017-10-24
Fix 5.x timeout

Godoy:
Fixes an error that doesn't allow to set the `timeout` option for more than 30 seconds.

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -87,11 +87,7 @@ async function runPa11yTest(url, options, state) {
 	// We use a state object which is accessible from the
 	// wrapping function
 	options.log.debug('Launching Headless Chrome');
-
-	const puppeteerConfig = options.chromeLaunchConfig;
-	puppeteerConfig.timeout = options.timeout;
-
-	const browser = state.browser = await puppeteer.launch(puppeteerConfig);
+	const browser = state.browser = await puppeteer.launch(options.chromeLaunchConfig);
 	const page = await browser.newPage();
 
 	// Intercept page requests, we need to do this in order
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -140,9 +140,11 @@ async function runPa11yTest(url, options, state) {
 
 	// Navigate to the URL we're going to test
 	options.log.debug('Opening URL in Headless Chrome');
-	await page.goto(url, {
-		waitUntil: 'networkidle'
-	});
+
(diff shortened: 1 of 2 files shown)


luizpedone:
👍

lib/pa11y.js
```diff
@@ -87,7 +87,11 @@ async function runPa11yTest(url, options, state) {
 	// We use a state object which is accessible from the
 	// wrapping function
 	options.log.debug('Launching Headless Chrome');
-	const browser = state.browser = await puppeteer.launch(options.chromeLaunchConfig);
+
+	const puppeteerConfig = options.chromeLaunchConfig;
+	puppeteerConfig.timeout = options.timeout;
+
+	const browser = state.browser = await puppeteer.launch(puppeteerConfig);
```
andrewmee: It feels a little bit odd to me to use the same timeout option to both set the Headless Chrome startup timeout as well as the page timeout.

I suggest we remove this part, so if someone wants to set a Chrome startup timeout they can still do that separately within the options.chromeLaunchConfig variable without it getting overridden by the page-specific loading (`goto`) timeout.

Godoy: Thanks for review @andrewmee! 

Yeah, that wasn't an easy decision. When I had taken this decision I was thinking that could be redundant for users need to pass two timeouts to pa11y. In a slow page case, I think that will be necessary to set both timeouts for effectively works.

Anyway, I'll change accordingly with your suggestion as soon as possible :)

Godoy: I've made the changes that you suggest.

Also I've understood that puppeteer timeout is only related to chromium launch, as I found in puppeteer docs:

`timeout <number> Maximum time in milliseconds to wait for the Chrome instance to start. Defaults to 30000 (30 seconds). Pass 0 to disable timeout.` 

Thx

Godoy:
@andrewmee done 😄

andrewmee:
Looks great, thanks @Godoy.

One interesting thing I didn't spot when first reviewing this is that the pa11y timeout option is also (already) used to provide a timeout on the entire promise return; https://github.com/pa11y/pa11y/blob/5.x/lib/pa11y.js#L47-L53

This will of course mean that the full timeout period will never be available to the `goto` command, as some of it will have been taken up in the overhead of launching chrome, etc.  We may need to improve the documentation slightly around this to make that clearer.

Still this is an important fix for people who were seeking to set timeout to be longer than puppeteer's default 30 seconds, so thanks very much!

PR #355, opened 2017-11-29
Add reporter, threshold and level as configuration options

andrewmee:
I think these are the last options that can be usefully set via the command line interface but not the configuration options.

Also documenting and adding a couple of tests to ensure the supremacy of CLI options over config options which seems to be the existing behaviour and does make sense to keep I think.


README.md
```diff
@@ -343,6 +343,8 @@ Configuration
 
 Pa11y has lots of options you can use to change the way Headless Chrome runs, or the way your page is loaded. Options can be set either as a parameter on the `pa11y` function or in a config file used by the command-line interface.
 
+When options can be set both in the configuration and using the command-line interface, it is the value set in the latter which will take priority.
```
hollsk: nitpicky in teh extreme, but i'd shorten the last clause of this sentence to "..., the value set in the latter will take priority".

bin/pa11y.js
```diff
@@ -118,15 +116,18 @@ async function runProgram() {
 }
 
 function processOptions(log) {
-	const options = extend({}, loadConfig(program.config), {
+	// CLI options take precedence over config options (which take precedence over defaults)
```
hollsk: useful commenting 👌

README.md
andrewmee: That's a great suggestion; I did think my wording there seemed a bit clunky.

rowanmanning:
Firstly, this is good and I'm glad we're adding this feature 🎉, but I'm afraid I'm about to drop a big comment on you because you were very brave and kindly decided to work on one of the less easy to understand parts of the codebase 😅 

There's some stuff in here that I _think_ breaks the way that defaults work, because all of these options have no use in the JavaScript API. I think in the docs we're also considering "config" and "options" to be the same thing, which isn't quite right. Gonna try and explain :) (explaining code is hard).

Before the review stuff I'd like to clarify the three different places where users can configure Pa11y, as this will make my review comments a little more clear (hopefully):

  1. As an **options** object. This is only available through the JavaScript API. The Command Line API builds its own options object based on the next two types of configuration
  2. As a **config** file. This is only available through the Command Line API. You can set all of the available **options** by specifying them in this file
  3. As flags on the command line (also only Command Line API)

So the only code that actually _uses_ `reporter`, `threshold`, and `level` is `bin/pa11y.js`. `lib/pa11y.js` doesn't care about these at all and, if you pass them via the JavaScript API, then no behaviour will change (which I guess you know as there are only options changes in there). This means that the documentation examples you give are incorrect, those options do nothing when `pa11y` is called and so could cause some confusion. The change here could be quite simple – maybe we just need to indicate that these options are only available via the CLI and change the example to be a JSON config object?

In this PR we're now extending `pa11y.defaults` in the CLI, but that means that the options are defaulted twice. I'm not sure if that's ideal, I can't think of a concrete example but it feels like there could be possible side-effects when setting some of the deeply nested Chrome options. Wondering whether we should remove these new options from `lib/pa11y.js` and just have them in `bin/pa11y.js` so you extend a new object with just those configurations.

Sorry for the long (and maybe rambling) feedback, do say if I've misunderstood/got it wrong. Very happy to merge this when:
1. documentation is corrected to indicate that the new options are only available in config files
2. _maybe_ there should be no changes to `lib/pa11y.js` or the unit tests as part of this PR, as the lib behaviour is not changed at all

andrewmee:
@rowanmanning thanks for the mega comment, I knew I was missing a few pieces with this (totally forgot I wanted to go back and see if I could address setting the defaults twice for example).

Oh I see what you mean about my documentation not making any sense!  I always forget the JS interface for pa11y, since I don't tend to use that one myself.  I just blithely copied the example format thinking there was parity between JS option setting and config/json option setting... Agreed we need to make the division here more explicit.

As an aside, I think the three ways of setting options and the fact that none of them quite line up is definitely a bit confusing... Agreed we need to do some more work in the readme so outline which options are available in each.

In the meantime I'll have a look again at `lib/pa11y` and see if I can neaten this up a bit!

rowanmanning:
Oh I just noticed another thing: the `default` flag seems to be always on. It's fine on the `5.x` branch. I _think_ this may be caused by the double extending

andrewmee:
I've tried to provide better clarity in that configuration section about our three ways to set options and how they related to each other, but would be great to hear some feedback around if that's improved things at all.

There's a slight temptation to detail the use of the CLI options on all the relevant configuration options, but it feels like it might make it all too verbose and confused.

PR #363, opened 2018-01-11
Adding a new action for redirecting to a specified url

stooobot:
We have a case where we are using the existing actions to log into an app by filling in fields and clicking submit buttons. Once logged in the app redirects to a landing page but this is not the page we want to run Pa11y against. There is another page, that you need to be authenticated to view (hence the logging in steps), that you can't get to directly from the landing page.

This new 'redirect' action e.g. "redirect to http://pa11y.org/" allows us to specify a new url to goto once our login actions have finished running.

Hopefully useful for others with similar circumstances...

### DIFF
--- a/README.md
+++ b/README.md
@@ -668,7 +668,8 @@ pa11y('http://example.com/', {
         'wait for fragment to be #page-2',
         'wait for path to not be /login',
         'wait for url to be https://example.com/',
-        'wait for #my-image to emit load'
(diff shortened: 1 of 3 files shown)


rowanmanning:
Sorry @stooobot, I forgot to also mention that we need to document the new action in the README. Let me know if you don't have time, then I can do it :)

PR #377, opened 2018-02-07
Support `options.page`

eloytoro:
This PR aims to achieve 2 main issues

1.- I noticed that if the actions part of the test fails the whole function will throw and the promise will be rejected, causing the process to end and close the browser, however its impossible to access the page at that point. This PR makes it so I can take screenshots of the page when errors are thrown because I would own the page.
2.- Also noticed that if the browser passed then the page created by the test will never close and would hoard memory. Causing long-lasting browser instances to keep these pages running forever.

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -64,6 +64,8 @@ async function pa11y(url, options = {}, callback) {
 	} catch (error) {
 		if (state.browser && state.autoClose) {
 			state.browser.close();
+		} else if (state.page && state.autoClosePage) {
+			state.page.close();
 		}
 
 		// Run callback if present, and reject with error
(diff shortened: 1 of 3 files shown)


rowanmanning:
Oh I was meaning to do this! Thanks, I'll take some time to review properly as soon as possible :)

eloytoro:
I didn't write a test case to check that the page closes after done testing, should I include that in this PR?

rowanmanning:
If you could, yeah that'd be great :)

eloytoro:
done!

rowanmanning:
Looks good, and I've tested locally :) thanks for doing this? I'll open a second PR adding the documentation because I think its complicated and differs from the rest of the option documentation.

PR #380, opened 2018-02-14
Add documentation for the page option

rowanmanning:
This also ensures that an error is thrown if you attempt to use the page
option without the browser option.

@eloytoro possibly you'd like to take a look at this?

### DIFF
--- a/example/puppeteer/index.js
+++ b/example/puppeteer/index.js
@@ -1,0 +1,64 @@
+// An example of running Pa11y programmatically, reusing
+// existing Puppeteer browsers and pages
+'use strict';
+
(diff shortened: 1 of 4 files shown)


example/puppeteer/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+	} catch (error) {
+
+		// Output an error if it occurred
+		console.error(error.message);
+
+		// Close the browser instance and pages if theys exist
+		if (pages) {
+			for (const page of pages) {
+				await page.close();
```
eloytoro: seems a bit redundant, closing a browser should also close all of the pages

Issue #428, opened 2018-10-04
Use Given Puppeteer Page Instance without page.goto()


kevinatown:
Hey! I think it will be great for a way to tell pa11y to ignore the URL and not run `page.goto()` and just use the provided puppeteer page instance. This would allow for:

- a single puppeteer page instance to be programmed separately using puppeteer's own api
- once a page instance manipulated/rendered it can be verified using e2e tests and passed to pa11y

It should be trivial to implement; adding an option like `ignoreUrl` that defaults to false and requires a page instance.

kevinatown:
Also I got a branch with what I was thinking and tests, it was super simple.

hollsk:
Cool! We'd be happy to review a PR if you'd like to submit it 😄

kevinatown:
@hollsk Thanks for your quick response, heres the PR. lmk what y'all think.

hollsk:
Thanks @kevinatown ! Haven't had time for a proper look yet, hoping to do that on friday. From a quick scan it looks great though 😄 

Will get back to you shortly!

kevinatown:
Awesome! Let me know if there is anything I need to change!

AndrewLeedham:
@hollsk Any movement on this. Would be a great addition.

hollsk:
Yes, I agree! I need somebody with a bit more experience than I have to review, and I've been struggling to find somebody who has the time. 

But I haven't forgotten about it, and I'm still on the lookout for a suitable reviewer, hoping to find somebody soon. Thanks for your patience 😄

kevinatown:
@hollsk small bump on the status of the review

thaffenden:
Would also love to use this if you have a chance to finish off the PR.

hollsk:
No promises on timelines, but we've managed to get a bit of momentum on our backlog now, so hopefully we'll get back to this one soon. 

Thanks for your patience, everyone - i know it's taken a while!

joeyciechanowicz:
Closing as the PR is merged

PR #429, opened 2018-10-05
Use given Puppeteer page instance

kevinatown:


kevinatown:
#428

joeyciechanowicz:
#377 added support for passing in a `page`. When you do, the URL will be ignored anyway.

The functionality of this PR seem achievable by either:

1. Don't pass a `page`: the URL is used
2. Pass a `page`: The URL is ignored

kevinatown:
Hey @joeyciechanowicz! Thank you so much for your answer. 

It looks like this functionality is not present in #377. #377 just adds the possibility to pass a puppeteer page instance. Pa11y still forces puppeteer to go to the provided URL (line 178 of lib/pa11y.js, which is before the code sniffer) even if a page instance was passed. This overrides the given puppeteer page instance -- either by re-rendering the already render state or going to a complete different state. I addressed the use case for this in #428. 

This PR addresses that by simply skipping line 178 of lib/pa11y.js -- leaving the given  puppeteer page instance as it was originally passed to pa11y.

Thanks again for your message and please let me know if there is anything else I can do to help move this along!

joeyciechanowicz:
You are absolutely right! ⭐️ 

The current functionality seems a bit counter intuitive to me. 
If you've gone to the bother of setting up a `page` object then it should use it _as is_, rather than navigating to the URL. So `ignoreUrl` isn't needed, it would be implicitly true if you passed in a `page`.

kevinatown:
I agree with you completely. I'd rather not have a flag and just expect pa11y to use the page as it's given. I just didn't want to make the decision and implement a breaking change (as I am assuming others are using it as is and expecting it to continue on this way). 

I am more than willing to change this code and remove the flag if it helps to move this PR on tho. 😅

joeyciechanowicz:
It's a very valid point that we don't want to make a breaking change.

Maybe a halfway solution between the two would be good. If added an overload of the main `pa11y` function to either accept `pa11y(url, options)` and also `pa11y(options)`.
When using the second, you have to provide a `browser` and `page`, and it will use the page as-is.

kevinatown:
good solution, I'll move around the code this weekend.

joeyciechanowicz:
@kevinatown In that case this seems like a great solution.

README.md
```diff
@@ -439,6 +439,18 @@ pa11y('http://example.com/', {
 
 Defaults to an empty array.
 
+### `ignoreUrl` (boolean)
+
+Whether to use the provided [Puppeteer Page instance][puppeteer-page] as is or use the provided url.
+
+```js
+pa11y('http://example.com/', {
+    ignoreUrl: true
```
joeyciechanowicz: Can you add `page` to show it's required to pass these two in when using `ignoreUrl`.
Otherwise this looks great 🎉 . Should be good to merge after that tiny change!

kevinatown: @joeyciechanowicz def! I'll try and get to it this week. Thanks!

kevinatown: @joeyciechanowicz added both `page` and `browser` to the config and merged in master. Lmk if there's anything else thanks!.

PR #434, opened 2018-10-18
specifies node v8.12 in .nvmrc facilitating npm ci installs

jpw:
Specifies a minor version of node which ships with a version of npm that supports the `ci` argument. Going forward, I suggest developers should install dependencies using `npm ci` which guarantees reproducible builds. This is because `npm ci` uses the `package-lock.json` file as the source of authority for all dependencies. This PR also includes related changes to the lockfile.


jpw:
(Before today there was no `.nvmrc` but yet a lockfile in git, which would mean constant changes to the lockfile as devs installed & pushed, and unpredictable builds.)

rowanmanning:
I'm not sure we want to include an `.nvmrc` file, do we? Not everybody uses nvm, and this only actually does anything if you run `nvm use` before you try to install anything or run Pa11y in development.

My process just now:
1. Pull changes from upstream
2. Run npm install
3. Get `package-lock.json` changes because I'm running Node.js 10 locally

jpw:
>2. Run npm install

Yep, one will almost certainly get `package-lock.json` conflicts if not running [`npm ci`](https://docs.npmjs.com/cli/ci) to install dependencies. Checking in `package-lock.json` to git does kinda require `npm ci` to be used, to avoid constant conflicts in the `package-lock.json` caused by the differences in how `npm i` behaves across different versions of node & `npm`.

Before using `.nvmrc`, the problem I hit was: I checked out pa11y and ran the tests which immediately exploded because my default node is v6. Hollie had similar problems IIRC.  But in `package.json`:
```
  "engines": {
    "node": ">=8"
  },
```
In other recent news, [Node.js 8.12.0 ships with a version of `npm` that supports `npm ci`](https://nodejs.org/en/download/releases/), and I'm liking `npm ci` for reasons described in the [Springer Nature Front-end playbook](https://github.com/springernature/frontend-playbook/blob/master/practices/managing-node-projects.md#specifying-versions-of-node) :)

So I am currently favouring using `nvm` in conjunction with `npm ci` to _enforce_ reproducible builds and thereby reduce "but it works/doesn't work on my machine" problems. That's my > 2p ;)

jpw:
...but if we don't want the nvmrc file will remove ;)

jpw:
I'd like us to have a think about this a bit more, as a previous commits added the `.nvmrc` file as well as a `package-lock.json` -- so they are going to be disruptive (constant diffs in the package lock) unless we can agree to either remove both those things, or everyone use `npm ci`.

So I think it would be good to make a decision either way.

I'm a noob here & don't want to rock the boat, and have nothing but hugs for pa11y contributors, so am genuinely happy to rm both files 🙂  I like `npm ci` though, and will summarise here what I write about in more length in the [SN Playbook on managing node projects](https://github.com/springernature/frontend-playbook/blob/master/practices/managing-node-projects.md);

1. predictable installs.
2. much faster install if no `./node_modules` dir present, e.g. on CI.
3. the lock file enables dep management tools to flag (and in some cases, fix) outdated deps deep in the dep graph (e.g. `snyk`, [`npm audit fix`](https://blog.npmjs.org/post/174001864165/v610-next0)). This means we can patch without having to wait for every author to publish new packages down that branch of the graph. This is a big win IMO. Not only for the reasons of security & warm fuzzies, but it prevents lots of noise from dep analysis tools (e.g. github) for our users.
4. bash & zsh shell extensions exist for auto-reading nvmrc files, so no need to remember to run `nvm use` once set up. I suspect they do have a slight performance impact though when opening sessions, but I work across a variety of node projects most days, so it's pretty much essential for me.

Downsides I think are:
1. initial hurdle.
2. constant conflicts in the lock file if ppl don't `nvm use`. On the upside, this is a very visible flag something has gone wrong.
3. more things I can't think of? I think @josebolos may have some ideas.

josebolos:
I've been trying to do some research but I can't seem to be able to find any authoritative source that shows the pros and the cons of each approach, or talks about best practices. To be honest, I'm finding difficult to understand the implications of each approach.

At the moment, we (developers) are currently specifying supported node versions in at least three different ways:

* In the `engines` field of the `package.json` file
* In the travis.yml file
* In the .nvmrc file

This opens the door to inconsistencies between them. Particularly, if my understanding of the travis docs are correct this may backfire, as having an .nvmrc file means that travis will ignore the versions specified in the travis.yml file (correct me if I'm wrong). This may problematic in many projects as we tend to run the tests in more than one version (all stable/LTS versions, usually).

I would also like to have a bit more of thought about how these kind of changes could affect:

* Users installing pa11y as a cli tool (e.g. `npm install -g pa11y)
* Developers using pa11y as an npm module
* Deployments to PaaS systems (Heroku, Cloudfoundry, etc.) of both the pa11y cli or a node app using the pa11y as an npm module
* Developers working on pa11y itself

The fact that `npm ci` is also mentioned it's not helping me, I think that my brain is still on holiday.

jpw:
I knew you would have ideas @josebolos 🙂  and thanks for having a think.

Yes it is tricky, and I am not 100% sure either. Here's my understanding of the first points you raise.

1. [ `engines`](https://docs.npmjs.com/files/package.json#engines). This seems to be advisory only. It warns when installing _dependencies_ that have a semver mismatch with the node version specified by `engines`. (There used to be [`engineStrict`](https://docs.npmjs.com/files/package.json#enginestrict) which would I think fail, but that was removed in npm 3.)  It is true that other software can parse the `engines` field and do things appropriately, e.g. CI software (such as CloudFoundry node buildpack) so it might be useful for a particular infrastructure. It seems to have no effect on the version of _node_ the developer is actually using without additional software. So to be clear I am not advocating removal of the `engines` field as users may well be relying on it in their architecture.

1. travis. This is not a problem because:
> The .nvmrc file is only read when node_js key in your .travis.yml files does not specify a nodejs version
[source](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions-using-nvmrc)

(So if only testing on one node version, using an `.nvmrc` _could_ simplify your travis config, but am not recommending that here of course.)

You are right in that maintaining node versions in more than one place is not DRY, but I think it's also right `engines` and `.nvmrc` are used for subtly different things.

re. `nvs` I was not aware of that. There's a recent [open issue](https://github.com/jasongin/nvs/issues/113) suggesting it could read any `.nvmrc` file which seems to me to make sense. From that issue it looks like we could add a `.node-version` which would seem to accomplish the same thing...

--- 

Your subsequent  thoughts on the various ways pa11y is used I am _much_ less clear on and lack experience thinking about this stuff, so am open to suggestions. I will try to think some more about them...

joeyciechanowicz:
This was recetnly discussed in the pa11y monthly meeting.
https://pa11y.slack.com/archives/CK7FLLZRU/p1561369593006900

There was some conensus to specifying `"node": ">=8"` in the `engines` section rather than using an `.nvmrc`

PR #449, opened 2019-04-01
Seems like Mocha v6 can be used without any changes

paazmaya:
Did not commit `package-lock.json` since it seemed to have so many other changes...


paazmaya:
Trying with something small to get started contributing.
Are there anyone else besides @rowanmanning who could do reviews and merges, as I assume he is busy..?

Issue #441, opened 2019-01-04
Set cookie with `headers.Cookie` doesn't work and break the page


radswiat:
## Expected behaviour

Pa11y should set cookie on the browser.
Pa11y should be able to load page with `headers.Cookie` without need to specify `headers.Accept`

## Actual behaviour

Pa11y is not setting cookie on the browser ( debugged with chromium dev tools )
Pa11y is failing to load page with 404 error when setting `headers.cookie` without also specifying `Accept` with ` 'text/html, application/ xhtml + xml, application/ xml; q = 0.9,*/*;q=0.8',`

## Steps to reproduce

```
     import pa11y from 'pa11y'

      pa11y('http://localhost:8000/test', {
        headers: {
          // puppeteer returns 404 without `Accept` header once `Cookie` is set
          Accept: 'text/html, application/xhtml + xml, application/xml;q=0.9,*/*;q=0.8',
          // Cookie is not being set on the browser
          Cookie: 'basketId=123',
        },
    })
```

## Environment:

Pa11y:      5.1.0
Node.js:    8.11.2
npm:        5.6.0
OS:         18.2.0 (darwin)

PR #458, opened 2019-04-25
Exclude Cookie header from being lowercased

liamkeaton:
The headers keys are being lowercased before being set, it's not clear
why but this is causing the Cookie header to not be set correctly in
puppeteer.

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -143,7 +143,11 @@ async function runPa11yTest(url, options, state) {
 				options.log.debug('Setting request headers');
 				overrides.headers = {};
 				for (const [key, value] of Object.entries(options.headers)) {
-					overrides.headers[key.toLowerCase()] = value;
+					if (key === 'Cookie') {
+						overrides.headers[key] = value;
+					} else {
+						overrides.headers[key.toLowerCase()] = value;
+					}
 				}
 
 				// Override the request POST data if present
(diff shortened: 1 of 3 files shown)


thibaudcolas:
@liamkeaton could you explain what you mean by "not be set correctly"? This seems to be working fine for me at least.

From what I can see in the Puppeteer docs (https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#requestheaders), it seems like it's expected for header names to be lower-case. It doesn't say why though 😅.

liamkeaton:
@thibaudcolas by not correctly I meant not at all. Before this change I had a session cookie that needed to be set for the page to be accessed. I did find it weird why this hadn't been spotted sooner though if this was the case!

thibaudcolas:
👍 gotcha. This is working fine for me at least (https://github.com/thibaudcolas/wagtail-tooling/blob/8663e51c2a395688e6fefd9e3156d5b5ae92bff8/accessibility/pa11y-test.js#L93-L107), so there might be something else at play here. Let's see what other people think.

liamkeaton:
@thibaudcolas That is very strange, without this fix it fails and with it passes. There is also an [issue](https://github.com/pa11y/pa11y/issues/441) that seems to be similar to what i'm experiencing.

liamkeaton:
I put together a [minimal use case](https://github.com/liamkeaton/pa11y-cookie) to get to the bottom of this and I disproved my theory. The cookie is getting parsed correctly whether upper or lower cased when using pa11y directly.

liamkeaton:
I've found this is a concurrency issue with Pa11y-CI, will issue a PR over there to correct it.

Issue #448, opened 2019-04-01
Update to recent HTML_CodeSniffer and use it as a dependency


paazmaya:
## Expected behaviour

Have a recent version of `HTML_CodeSniffer` used in this project.

## Actual behaviour

Using a slightly older version, which is manually copied and comitted in this project code.

## Steps to reproduce

Looking to `lib` folder and in `package.json`...

## Environment:

```
Pa11y:      5.1.0
Node.js:    8.15.1
npm:        6.9.0
OS:         18.5.0 (darwin)
```

paazmaya:
Shall I create PR for this?

hollsk:
Yes please, but I think there's a reason we do it in the weird way that we do (we _used_ to have it as a dependency). I don't remember what the reason is, though. 

@rowanmanning what was the reason for not using HTML_CS as a normal dep?

paazmaya:
Perhaps because it (`HTML_CodeSniffer`) did not provide build version of itself in its npm package...

https://github.com/squizlabs/HTML_CodeSniffer/releases/tag/2.3.0

hollsk:
Haha! Yes! That _is_ why - I was thinking to myself "... it's something to do with the build, isn't it?" 

But yes, that's the reason. Good catch!

paazmaya:
CLosing this since the PR #460 is merged.
Might fix #420 too...
Would it make sense to release a new version now?

Issue #335, opened 2017-10-04
 HTML CodeSniffer: undefined error


mncrff:
When trying to run a pa11y scan through the command line on any site, I get the following error:

```
Error: HTML CodeSniffer: undefined is not an object (evaluating 'window.HTMLCS.process')
    at Object.<anonymous> (/usr/local/lib/node_modules/pa11y/lib/pa11y.js:71:16)
    at Object.f [as onCallback] (/usr/local/lib/node_modules/pa11y/node_modules/once/once.js:25:25)
    at /usr/local/lib/node_modules/pa11y/node_modules/node-phantom-simple/node-phantom-simple.js:636:30
    at Array.forEach (native)
    at IncomingMessage.<anonymous> (/usr/local/lib/node_modules/pa11y/node_modules/node-phantom-simple/node-phantom-simple.js:617:17)
    at emitNone (events.js:110:20)
    at IncomingMessage.emit (events.js:207:7)
    at endReadableNT (_stream_readable.js:1045:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

Any help is appreciated!

rowanmanning:
Hi @mncrff. Could I get the following information from you? (You can get most of this by running `pa11y --environment`):

- Pa11y version
- Node.js version
- OS Version
- An example of a page that fails

nzambello:
Hi everyone, I'm new of pa11y.
I had the same issue and I resolved adding `-H http://squizlabs.github.io/HTML_CodeSniffer/build/HTMLCS.js` as described [here](https://github.com/pa11y/pa11y#htmlcs-string)

mncrff:
@rowanmanning, here is the information:

Pa11y:      4.13.0
Node.js:    8.2.1
npm:        5.3.0
PhantomJS:  2.1.1
OS:         16.7.0 (darwin)

A page that fails: https://www.microsoft.com/en-us/

Oddly enough I had a few other pages that were having the same error yesterday, but not today =/

mncrff:
Wow, @nzambello 's suggestion seems to work. Is there a reason why `http://squizlabs.github.io/HTML_CodeSniffer/build/HTMLCS.js` works but `lib/vendor/HTMLCS.js` doesn't? Why do they differ? Without knowing anything about the inner workings of pa11y, the `lib/vendor/HTMLCS.js` seems buggy.

Godoy:
[...]

PR #460, opened 2019-05-16
Use html_codesniffer from npm as oppose to including it

paazmaya:
Locally could not pass the integration tests..
..and seems to be the same with Travis.
Could guess that some might be due to possible new accessibility checks but then some seemingly random failures..
#448


paazmaya:
Rechecked with clean install, the `npm test` tests pass.
There must be something changed in the behaviour of `html_codesniffer` between version 2.2.0 and 2.3.0...

hollsk:
A memory stirs...

I seem to recall I had this same problem when I tried to do what you're doing now, so I abandoned the attempt and went back to including htmlcs with the built javascript at http://squizlabs.github.io/HTML_CodeSniffer/build/HTMLCS.js instead 🤔

hollsk:
Oh, hm. Looks like I was maybe thinking of [this issue](https://github.com/pa11y/pa11y/issues/335), which was something different (it caused PhantomJS to crash, instead of just some tests failing).

paazmaya:
Just recording this diff here https://github.com/squizlabs/HTML_CodeSniffer/compare/2.2.0...2.3.0

paazmaya:
Also noting down for myself the command for running single test file:

```sh
npx mocha test/integration/setup.test.js test/integration/cli/standard.test.js --timeout 5000
```

paazmaya:
A test case at Travis seems to have timed out. Could anyone re-run it?

hollsk:
Done! @joeyciechanowicz it's all yours for review 😉

test/integration/cli/config.test.js
```diff
@@ -50,7 +50,7 @@ describe('CLI config', () => {
 		// the page title, so reading the title confirms that headers were sent by Pa11y
 		it('tests the page using the specified HTTP method', () => {
 			assert.isArray(pa11yResponse.json);
-			assert.lengthEquals(pa11yResponse.json, 1);
+			assert.lengthEquals(pa11yResponse.json, 13);
```
joeyciechanowicz: Theres a lot of tests cases in here where the numbers have increased significantly. Is that a result of `HMTLCS.js` picking up more problems now? Or are we now not passing options to it correctly so we're getting back different results.

test/integration/cli/exit-codes.test.js
```diff
@@ -127,6 +127,7 @@ describe('CLI exit codes', () => {
 
 	describe('when the `level` config is set to "warning" but the `--level` flag is set to "notice"', () => {
 
+		/* Should add many more rules to ignore..
```
joeyciechanowicz: Is there a reason this test case it commented out? I think we should add the rules to ignore so we have a working test

test/integration/cli/exit-codes.test.js
```diff
@@ -172,6 +174,7 @@ describe('CLI exit codes', () => {
 
 	describe('when the `--level` flag is set to "notice"', () => {
 
+		/* Should add many more rules to ignore..
```
joeyciechanowicz: Is there a reason this test case it commented out? I think we should add the rules to ignore so we have a working test

test/integration/cli/reporter-csv.test.js
```diff
@@ -20,10 +20,10 @@ describe('CLI reporter CSV', () => {
 
 		it('outputs issues in CSV format', () => {
 			const lines = pa11yResponse.output.trim().split('\n');
-			assert.lengthEquals(lines, 5);
+			assert.lengthEquals(lines, 18);
 			assert.strictEqual(lines[0], '"type","code","message","context","selector"');
 			lines.slice(1).forEach(line => {
-				assert.match(line, /^"(error|warning|notice)","[^"]+","[^"]+",".*","[^"]+"$/i);
+				assert.match(line, /^"(error|warning|notice)","[^"]+","[^"]+",(".*"|null),"[^"]*"$/i);
```
joeyciechanowicz: The CSV output shouldn't contain `null` values. We might need to add some pre-formatting to convert null to empty strings or `\n` etc

test/unit/lib/pa11y.test.js
```diff
@@ -122,7 +122,7 @@ describe('lib/pa11y', () => {
 
 		it('loads the HTML CodeSniffer JavaScript', () => {
```
joeyciechanowicz: It might be useful to include a test that checks the file is included correctly when `pa11y` is installed as a dependency, rather than used directly.
I'll have to have a think about how that could be done.

test/integration/cli/config.test.js
paazmaya: More issues being found by html_codesniffer

test/integration/cli/exit-codes.test.js
paazmaya: Too long list in my opinion and maintaining it later...

test/integration/cli/exit-codes.test.js
paazmaya: Too long list in my opinion and maintaining it later...

lib/pa11y.js
```diff
@@ -193,7 +193,7 @@ async function runPa11yTest(url, options, state) {
 	// We only load these files once on the first run of Pa11y as they don't
 	// change between runs
 	if (!htmlCodeSnifferJavaScript) {
-		htmlCodeSnifferJavaScript = await fs.readFile(`${__dirname}/vendor/HTMLCS.js`, 'utf-8');
+		htmlCodeSnifferJavaScript = await fs.readFile(path.resolve(`${__dirname}/../node_modules/html_codesniffer/build/HTMLCS.js`), 'utf-8');
```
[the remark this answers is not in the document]
paazmaya: Thanks! I simply changed this to make test case pass, which wanted the full filepath.

test/integration/cli/reporter-csv.test.js
paazmaya: Should that be done in a different PR, but before next release?

test/unit/lib/pa11y.test.js
paazmaya: Could that happen in a separate PR?

test/integration/cli/reporter-csv.test.js
joeyciechanowicz: My worry that is that this will introduce a breaking change for those who rely on the CSV output.

test/unit/lib/pa11y.test.js
joeyciechanowicz: I'd say that we are actually ok here. As all we would be testing is functionality of `require.resolve` which we should assume works correctly 😏

test/integration/cli/reporter-csv.test.js
paazmaya: Looks like csv reporting is generated in another project, namely `pa11y-reporter-csv`. How to solve?

joeyciechanowicz: Without delving into `pa11y-reporter-csv`, I would hazard a guess that for notices the `context` is now `null` and it doesn't print that out nicely.
We can either `map` our results to use an empty string rather than `null`, or update the reporter to handle `null` values instead.

paazmaya:
@joeyciechanowicz how to proceed?

joeyciechanowicz:
If you use slack, then there is a Pa11y slack we use to discuss technical issues. Details can be found here: http://pa11y.org/contact/

I'm working on a branch to help with this. I'm thinking that we want to rejiggle the tests so they don't rely on notices (which there are now tons of) and instead use warnings which are more explicit.

joeyciechanowicz:
@paazmaya I've opened a PR to your fork with some changes that address the issues on this PR https://github.com/paazmaya/pa11y/pull/1

test/integration/cli/reporter-csv.test.js
joeyciechanowicz: I've looked at this and it just seems that previously there was nothing that would have a `null` context. Theres a test case for `context` being null as well. So this is desired

Issue #461, opened 2019-06-13
Event listeners do not fire if prototype listener is overridden


kkoskelin:
## Expected behaviour
When I use an action like `set field #name to Jane Smith`, I expect my `onChange` function to fire when set within a React app as such:
`<input type="text" id="name" defaultValue="" onChange={(e) => console.log("value has changed")}>`

This can be important if the change event causes other form elements on the page to be changed or revealed.

## Actual behaviour
onChange function assigned through the React app does not fire.

This might also be related to other frameworks that use shadow DOM. Related to #384 too?

## Steps to reproduce

Try the fiddle at [https://jsfiddle.net/kkoskelin/j2ry1vLe/1/](https://jsfiddle.net/kkoskelin/j2ry1vLe/1/)

The "direct" setter method (first button) will access the setter for 'value' as if it were an attribute.  The 'prototype' setter method accesses the setter on the element's prototype before dispatching the event which DOES trigger `onChange` events on text inputs.

Reference: [https://github.com/pa11y/pa11y/blob/5.x/lib/action.js#L96](https://github.com/pa11y/pa11y/blob/5.x/lib/action.js#L96)

## Environment:
```
Pa11y:      5.1.0
Node.js:    10.15.3
npm:        6.4.1
OS:         18.6.0 (darwin)
```

joeyciechanowicz:
Fixed with #462 merged

josebolos:
Fix released as part of [version 5.2.0](https://github.com/pa11y/pa11y/blob/master/CHANGELOG.md#520-2019-07-04).

PR #462, opened 2019-06-17
Use prototype setter

kkoskelin:
Addresses issue #461 
Comment for maintainers: It had occurred to me to completely replace the `target.value = value` with the prototype value setter method within the actions file, but I was not 100% certain whether all elements would necessarily have a setter on their prototype. If you have greater confidence, one use could be replaced with the other and additional tests removed.


christophersbarrett:
👍

Issue #470, opened 2019-07-16
Correct URL for shields.io node version badge


josebolos:
The node version badge in the README currently points to

`https://img.shields.io/badge/node.js%20support-8-brightgreen.svg`

This hardcodes the version to 8, and requires an update every time that support changes.

We should change the badge's URL so we use the version that automatically figures out the supported versions.

PR #472, opened 2019-07-23
docs: update the badge URL for the Node.js version

sjparkinson:
Uses the dynamic URL that looks up the supported Node.js versions on npm.

Resolves https://github.com/pa11y/pa11y/issues/470.


sjparkinson:
The new badge will look like the following:

image

Vs. the current:

image

PR #485, opened 2019-09-12
Add some JSDoc to bin/pa11y.js

josebolos:
There's a bunch of functions on this file, so now at least it will be easier to understand what each one of them is doing, as there are no unit tests for them.

### DIFF
--- a/bin/pa11y.js
+++ b/bin/pa11y.js
@@ -12,6 +12,10 @@ const semver = require('semver');
 configureProgram();
 runProgram();
 
+/**
+ * Parse the flags and arguments passed to the CLI
+ * @returns {void}
+ */
 function configureProgram() {
 	program.version(pkg.version)
 		.usage('[options] <url>')
@@ -95,6 +99,10 @@ function configureProgram() {
 	program.url = program.args[0];
 }
 
+/**
+ * Test the page and generate the results
+ * @returns {void}
+ */
 async function runProgram() {
 	if (program.environment) {
 		outputEnvironmentInfo();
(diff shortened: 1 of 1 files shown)


bin/pa11y.js
```diff
@@ -124,9 +132,14 @@ async function runProgram() {
 	}
 }
 
+/**
+ * Process options based on precedence rules.
+ * CLI options take precedence over config options (which take precedence over defaults)
+ * @returns {Object} An object containing all the options that have been specified
```
joeyciechanowicz: You could do a `@typedef` and define the full config object....if you _really_ cared... 🤔

bin/pa11y.js
```diff
@@ -203,6 +241,14 @@ function requireFirst(stack, defaultReturn) {
 	}
 }
 
+/**
+ * Calculates if there are any errors to be reported (i.e. above the threshold)
+ * This is used for example to return the correct status code on the CLI
+ * @param {String} level - The level of reporting required (warning, notice, etc.)
+ * @param {Object} issues - The issues found by the test
```
joeyciechanowicz: The parameter is named `results`
```suggestion
 * @param {Object} results - The results found by the test
```

bin/pa11y.js
```diff
@@ -224,10 +270,21 @@ function isWarningOrError(result) {
 	return (result.type === 'warning' || result.type === 'error');
 }
 
+/**
+ * Create an array from a semicolon-separated string of options
+ * @param {String} val - The string of user-specified values
+ * @param {Array} array - The previous value, if any
```
joeyciechanowicz: Needs to be plural
```suggestion
 * @param {Array} array - The previous value(s), if any
```

bin/pa11y.js
josebolos: I don't! :trollface:

PR #494, opened 2019-09-27
Adds sanity checks against real sites

joeyciechanowicz:
We've had problems recently with new versions of pa11y breaking on some sites (cough cough AMD).

This is a small set of integration tests that ensure pa11y will run without throwing an exception against a set of sites, some of which we know have caused isues.


joeyciechanowicz:
Decied that this will make our tests flakey than they need to me.

PR #497, opened 2019-10-01
Update dependencies

paazmaya:
Simple as that..

Perhaps could look into adding https://renovatebot.com/ on all the @pa11y projects?

PR #498, opened 2019-10-05
Add renovatebot config for automated dependency updates

paazmaya:
This PR adds a configuration file for https://renovatebot.com/, which can be used once that service has been installed for this project.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,10 @@
+{
+  "extends": [
+    "config:base"
+  ],
+  "commitBody": "Signed-off-by: {{{gitAuthor}}}",
+  "automerge": true,
+  "major": {
+    "automerge": false
+  }
+}


josebolos:
Hi @paazmaya 

Thanks for creating the PR! After discussing with some of the core contributors, I think that we prefer a more manual approach to dependency handling. There are several reasons for this.

First, this project is completely run by volunteers, and unfortunately we can dedicate as much time to it as we would like to. We've had instances in the past where updating dependencies has caused breakage that wasn't be caught by tests, and due to the lack of time, a "broken" version of pa11y has stayed in `master` for too long, or has even been released.

We have a current example of a dependency breaking things: `nyc` is currently pinned at version 13.2.0 in pa11y, as updating to 13.3.0 or greater causes coverage to not be calculated for all the source files.

Lastly, we also do releases a few times per year, so it also feels more efficient with our time to bump the dependencies once before doing a new release, rather than updating dependencies every few days, for no particular benefit, and without being able to release the updates.

That said, we do like, whenever possible and as far as time allows, to bump dependencies that may contain an exploitable security vulnerability.

Thanks again for the PR and for all your contributions, but we won't be merging this at this time! I hope that's ok.

paazmaya:
Its ok, I still ❤️ and 🎧 to other PRs...

paazmaya:
Well #497 updates `nyc` and at least automation does not have any problems with it

Issue #483, opened 2019-09-10
Improve `environment` flag to help with debugging


josebolos:
I suggest outputting something like this, in addition to the version of pa11y itself. This could also be useful when debugging tests run using the library (e.g. from pa11y-dashboard).

```
System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
    Memory: 3.64 GB / 16.00 GB
  Binaries:
    Node: 8.15.0 - /usr/local/opt/node@8/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/opt/node@8/bin/npm
```

The above is an example output of [envinfo](https://www.npmjs.com/package/envinfo).

PR #509, opened 2020-02-24
Improve environment flag

ryhinchey:
Hi there! I've been using this project lately and decided to start contributing.  This PR is in reference to issue https://github.com/pa11y/pa11y/issues/483.

### DIFF
--- a/bin/pa11y.js
+++ b/bin/pa11y.js
@@ -3,6 +3,7 @@
 
 const buildReporter = require('../lib/reporter');
 const extend = require('node.extend');
+const envinfo = require('envinfo');
 const path = require('path');
 const pkg = require('../package.json');
 const program = require('commander');
@@ -117,7 +118,8 @@ function handleHelp() {
  */
 async function runProgram() {
 	if (program.environment) {
-		outputEnvironmentInfo();
+		const info = await outputEnvironmentInfo();
+		console.log(info);
 		process.exit(0);
 	}
 	handleHelp();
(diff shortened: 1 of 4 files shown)


bin/pa11y.js
```diff
@@ -298,22 +300,14 @@ function collectOptions(val, array) {
...
-	} catch (error) {}
-
-	console.log(`Pa11y:      ${versions.pa11y}`);
-	console.log(`Node.js:    ${versions.node}`);
-	console.log(`npm:        ${versions.npm}`);
-	console.log(`OS:         ${versions.os} (${process.platform})`);
+	return envinfo.run({
+		System: ['OS', 'CPU', 'Memory'],
+		Binaries: ['Node', 'Yarn', 'npm'],
+		Browsers: ['Chrome', 'Firefox', 'Safari'],
```
joeyciechanowicz: We don't need the `Browsers` section as chromium is provided by `puppeteer`.

bin/pa11y.js
```diff
@@ -298,22 +300,14 @@ function collectOptions(val, array) {
...
+		npmPackages: ['pa11y', 'pa11y-ci', 'puppeteer', 'pa11y-runner-axe']
```
joeyciechanowicz: This should just be `pa11y` IMO. The others can be found easily enough by checking the version on GitHub

bin/pa11y.js
```diff
@@ -298,22 +300,14 @@ function collectOptions(val, array) {
...
-	try {
-		versions.npm = require('child_process').execSync('npm -v').toString().trim();
-	} catch (error) {}
-
-	console.log(`Pa11y:      ${versions.pa11y}`);
-	console.log(`Node.js:    ${versions.node}`);
-	console.log(`npm:        ${versions.npm}`);
-	console.log(`OS:         ${versions.os} (${process.platform})`);
+	return envinfo.run({
+		System: ['OS', 'CPU', 'Memory'],
```
joeyciechanowicz: `'Shell'` would probably be useful.

```suggestion
		System: ['OS', 'CPU', 'Memory', 'Shell'],
```

bin/pa11y.js
```diff
@@ -117,7 +118,8 @@ function handleHelp() {
  */
 async function runProgram() {
 	if (program.environment) {
-		outputEnvironmentInfo();
+		const info = await outputEnvironmentInfo();
```
joeyciechanowicz: Can you add the `console.log` to `outputEnvironmentInfo` rather than having it in here.

joeyciechanowicz:
Thanks so much for contributing! Good issue to pick up.

bin/pa11y.js
ryhinchey: sounds good. if others are needed down the road they can be easily added

PR #536, opened 2020-10-29
Fix ignore url handling

### LINKED ISSUE #535: When setting a default of `ignoreUrl: true`, the `pa11y()` function shouldn't require a url parameter
**Please note:** This repository's issues are reserved for feature requests, bug reports, and other issues with the Pa11y library itself.

If you need help integrating Pa11y with your application, we recommend using [Stack Overflow](https://stackoverflow.com/questions/tagged/pa11y).

For a bug report, please use the template below:

---

## Expected behaviour
1) `options.page` exists, but there is no default `pa11y.default.page` is not defined. This is not a huge deal, it's just a little confusing, as `pa11y.default.browser` is defined as `null`.

2) When setting my own `pa11y.defaults.page`, I expected to be able to set `pa11y.defaults.ignoreUrl` and not have to provide the `url` parameter or `options.url`.

Additionally, if `pa11y.defaults.url` is set, then there is no need for any parameters to be passed to `pa11y()`.

## Actual behaviour
`pa11y()` expects a url, either as a param or via options. The code that handles this does not consider if `ignoreUrl` is in use. https://github.com/pa11y/pa11y/blob/master/lib/pa11y.js#L71

This results in me having to do the following if I pass in my own page object.

```js
// page defined elsewhere
pa11y.defaults = {
  ...pa11y.defaults,
  page,
  ignoreUrl: true,
};
await pa11y('');
```

The empty string passed in `await pa11y('');` seems like a really minor deal, but the setup and configuration of pa11y is separate from its usage in my code. I'm afraid it results in being confusing for others who might not understand why they need to use `pa11y('')` instead of `pa11y()`.

## Steps to reproduce
```js
// page defined elsewhere

pa11y.defaults = {
  ...pa11y.defaults,
  page,
  ignoreUrl: true,
};
await pa11y();
```

## Environment:
```
Pa11y:      5.3.0
Node.js:    12.18.2
npm:        6.14.8
OS:         19.6.0 (darwin)
```

wesrice:
Closes #535

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -34,6 +34,11 @@ async function pa11y(url, options = {}, callback) {
 		// Verify that the given options are valid
 		verifyOptions(options);
 
+		if (!url && !options.ignoreUrl) {
+			// eslint-disable-next-line max-len
+			throw new Error('A url must be defined via the `url` parameter or `options.url`, or `options.ignoreUrl` must be set to true.');
+		}
+
 		// Call the actual Pa11y test runner with
 		// a timeout if it takes too long
 		pa11yResults = await promiseTimeout(
@@ -58,7 +63,7 @@ async function pa11y(url, options = {}, callback) {
 /**
  * Parse arguments from the command-line to properly identify the url, options, and callback
  * @private
- * @param {String} url - The URL to run tests against.
+ * @param {(String|Object)} url - The URL to run tests against, or options object.
  * @param {Object} [options={}] - Options to change the way tests run.
  * @param {Function} [callback] - An optional callback to use instead of promises.
  * @returns {Array} the new values of url, options, and callback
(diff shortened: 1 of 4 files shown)


joeyciechanowicz (approved):
This is great, and some lovely tests. Just needs the merge conflicts fixed, which seems not too bad, and we're good to go 🚀

danyalaytekin:
Hey @wesrice, thanks for this contribution. `pa11y:master` has actually been renamed to `pa11y:main`, rather than deleted.  Thanks for the jumpscare, GitHub.

It looks like the repo forwarding from fork to origin doesn't work (which goes into my collection as the second example of it not working, alongside a mention of `master` in a GitHub Actions workflow).

I haven't read the content of your PR yet but @joeyciechanowicz seemed excited about it so if it remains relevant please do recreate it, targetting `main`.

Issue #534, opened 2020-10-25
Feature request - support playwright browser instance


mickyginger:
At https://frontendmentor.io we've been using pa11y to run accessibility testing against uploaded solutions to the platform.

We also used puppeteer to take a screenshot of the page (and do various other bits and pieces). However we had quite a lot of rendering issues, so we switched to playwright.

We had been passing in the puppeteer browser instance to pa11y in order to optimize the process, but now that we've switched to playwright, we're no longer able to do that. We're now experiencing some timeouts on staging which we think are related to performance (we're not actually sure because the only error we get is a timeout error.)

I'd like to have a go at modifying the `lib/pa11y.js` file to handle a playwright browser instance, but I thought it might be worth running it by you first, incase you're already looking into it or have some reckons on the best approach.

josebolos:
Hi @mickyginger,

Thanks for creating this issue. At the moment, we're not keen on in migrating away from `puppeteer`, as it currently covers all our needs, and migrating away from it would add to the maintenance burden. We're already stretched as it is.

Luckily, it looks like `puppeteer` has added support to Firefox in their latest version, so it may be possible to achieve what you want using this feature:
https://github.com/puppeteer/puppeteer/blob/v8.0.0/docs/api.md#puppeteerconnectoptions

mickyginger:
Thanks for your reply @josebolos, I'll close this for now then.

venkatsaivikram:
<img width="786" height="755" alt="Image" src="https://github.com/user-attachments/assets/a915f5b3-d6e0-4404-b14c-cb63616050fe" />

I had the same scenario, I'm using playwirght and wanted to integrate pa11y with it.. After debugging through pa11y code flow for generating report, got to know that it expects some puppeteer page object methods which are not available in playwright page object.. Luckily the above code works for me, in which the required props are overridden with noop functions . Might not be a standard way, but it does the job. Hope it helps.

PR #537, opened 2020-11-03
Replace puppeteer with playwright

mickyginger:
## Overview

This PR replaces Puppeteer with Playwright as the headless browser. This opens up the possibility to support multiple browsers.

## Approach

In this PR I have attempted only to replace Puppeteer with Playwright as simply as possible. I have not looked into refactoring or optimization.

## Notes

+ Unlike Puppeteer, Playwright makes use of [browser contexts](https://playwright.dev/#version=v1.6.2&path=docs%2Fcore-concepts.md&q=browser-contexts) to save memory. This is an avenue that might be worth investigating, although it might involve a reasonably large refactor.

+ I have only used Playwright's Chromium browser, however it should be reasonably trivial to pass an option to select Chromium, Firefox or WebKit. I think it's reasonable to only run tests against the Chromium browser given that the API should be identical for all three browsers. I'd be happy to look into this in a new branch if you do decide to introduce Playwright.

+ Playwright will wait 30000ms for elements to be attached to the DOM before timing out, which is the same as Pa11y's default timeout, this means that Pa11y will timeout before Playwright will exit when a DOM element cannot be found. I have set the default timeout to 5000ms, which seemed like a reasonable setting, but this could be set as an option.

+ I am getting the following warning when running certain tests:
  ```
  Transformation error for /Users/mickyginger/development/frontendmentor/pa11y/lib/pa11y.js ; return original code
  Expected opts.sync to be a function.
  ```
  which I have not been able to resolve, I believe it has something to do with using ES6 import notation in node, but I cannot see where I have introduced that anywhere in the project.

+ My build seems to be failing on node v12.19.1 on Travis, however I am not able to replicate this locally, so I have not been able to investigate.

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -7,7 +7,7 @@ const path = require('path');
 const {promisify} = require('util');
 const pkg = require('../package.json');
 const promiseTimeout = require('p-timeout');
-const puppeteer = require('puppeteer');
+const {chromium: playwright} = require('playwright');
 const semver = require('semver');
 
 const runnerJavascriptPromises = {};
@@ -49,7 +49,7 @@ async function pa11y(url, options = {}, callback) {
 			throw error;
 		}
 	} finally {
-		await stateCleanup(state);
+		await stateCleanup(options, state);
 	}
 	// Run callback if present, and resolve with pa11yResults
 	return callback ? callback(pa11yError, pa11yResults) : pa11yResults;
@@ -143,15 +143,20 @@ async function runPa11yTest(url, options, state) {
 /**
  * Ensures that puppeteer resources are freed and listeners removed.
  * @private
+ * @param {Object} options - Options to change the way tests run.
  * @param {Object} state - The last-known state of the test-run.
  * @returns {Promise} A promise which resolves when resources are released
  */
-async function stateCleanup(state) {
+async function stateCleanup(options, state) {
 	if (state.browser && state.autoClose) {
 		await state.browser.close();
 	} else if (state.page) {
-		state.page.removeListener('request', state.requestInterceptCallback);
-		state.page.removeListener('console', state.consoleCallback);
+		try {
+			state.page.removeListener('request', state.requestInterceptCallback);
+			state.page.removeListener('console', state.consoleCallback);
+		} catch (error) {
+			options.log.debug(error);
+		}
 		if (state.autoClosePage) {
 			await state.page.close();
 		}
@@ -179,7 +184,7 @@ async function setBrowser(options, state) {
 		// state object which is accessible from the
 		// wrapping function
 		options.log.debug('Launching Headless Chrome');
-		state.browser = await puppeteer.launch(
+		state.browser = await playwright.launch(
 			options.chromeLaunchConfig
 		);
 		state.autoClose = true;
@@ -199,7 +204,11 @@ async function setPage(options, state) {
 		state.page = options.page;
 		state.autoClosePage = false;
 	} else {
-		state.page = await state.browser.newPage();
+		state.page = await state.browser.newPage({
+			viewport: options.viewport,
+			userAgent: options.userAgent
+		});
+		state.page.setDefaultTimeout(5000);
 		state.autoClosePage = true;
 	}
 	// Listen for console logs on the page so that we can
@@ -209,10 +218,6 @@ async function setPage(options, state) {
 	};
 	state.page.on('console', state.consoleCallback);
 	options.log.debug('Opening URL in Headless Chrome');
-	if (options.userAgent) {
-		await state.page.setUserAgent(options.userAgent);
-	}
-	await state.page.setViewport(options.viewport);
 }
 
 /**
@@ -237,9 +242,6 @@ async function interceptRequests(options, state) {
 	if (!shouldInterceptRequests) {
 		return;
 	}
-	// Intercept page requests, we need to do this in order
-	// to set the HTTP method or post data
-	await state.page.setRequestInterception(true);
 
 	// Intercept requests so we can set the HTTP method
 	// and post data. We only want to make changes to the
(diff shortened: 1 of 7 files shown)


joeyciechanowicz:
Being able to support multiple browsers out the box would be great, and (despite the tests not passing) i'm suprised by how few changes this is requiring so far.

I'm all for this 👍 

>  It might be better to introduce a plugin pattern which would allow developers to build out their own headless browser plugins.
Though in principal this would be nice, I think the added cost in issues due to someone using a custom plugin would be overwhelming for this project. It would be better to just support one automation framework, but one that lets us use multiple browsers would be nice.

lib/pa11y.js
```diff
@@ -7,7 +7,7 @@ const path = require('path');
 const {promisify} = require('util');
 const pkg = require('../package.json');
 const promiseTimeout = require('p-timeout');
-const puppeteer = require('puppeteer');
+const {chomium: playwright} = require('playwright');
```
kkoskelin, line 10: *`chromium`

lib/pa11y.js
```diff
@@ -150,8 +150,12 @@ async function stateCleanup(state) {
...
 		await state.browser.close();
 	} else if (state.page) {
-		state.page.removeListener('request', state.requestInterceptCallback);
-		state.page.removeListener('console', state.consoleCallback);
+		try {
+			state.page.removeListener('request', state.requestInterceptCallback);
+			state.page.removeListener('console', state.consoleCallback);
+		} catch {
+			// do nothing
+		}
```
kkoskelin, line 158: Just need to provide an identifier for the caught error.  maybe log-debug, also?
```
} catch (error) {
  options.log.debug(error);
}
```

mickyginger: `options` are not being passed in here at present... Would I need to pass them into the `stateCleanup` method?

```
async function stateCleanup(options, state) {
  ...
}

async function pa11y(url, options = {}, callback) {
  ...
  } finally {
    await stateCleanup(options, state);
  }
}
```

kkoskelin: Quite right, that would be enough.

mickyginger:
Thanks for your input both... I've fixed up some of the specs, but I'm struggling with the cli specs. Can anyone point me in the right direction?

kkoskelin:
@mickyginger I've tried a bit, but the most I've gotten is 
`Error: Failed action: no element matching selector \"#unchecked-field\"`
(which failed during `test/integration/cli/actions-check-field.test.js`)

I've tried inspecting the page but I'm not familiar with playwright.  You might want to try altering your code to inspect the response to verify that the source of the page is actually getting loaded.  Right now I'm uncertain what it's getting.

mickyginger:
@kkoskelin When I run that spec I get the following error: 

```
  1) CLI action "check-field"
       when the field exists and is unchecked
         checks the field before running tests:

      AssertionError: 'object' === 'array'
      + expected - actual

      -object
      +array
```

`pa11yResponse.json` is returning `null` rather than an array, which I guess means that the stdout is not being captured correctly. Not sure where to look to resolve this though.

mickyginger:
I would be happy to pair on this if you're interested @kkoskelin?

kkoskelin:
> I would be happy to pair on this if you're interested @kkoskelin?

Let's take this conversation to the pa11y slack, if that's okay with you? [pa11y.slack.com](pa11y.slack.com)
Here's a hint of what I've found:
```
Error: Too many arguments. If you need to pass more than 1 argument to the function wrap them in an object.
    at Object.assertMaxArguments (/Users/kkoskelin/projects/pa11y/node_modules/playwright/lib/client/jsHandle.js:84:15)
    at Page.evaluate (/Users/kkoskelin/projects/pa11y/node_modules/playwright/lib/client/page.js:340:20)
    at Object.run (/Users/kkoskelin/projects/pa11y/lib/action.js:125:16)
    at runAction (/Users/kkoskelin/projects/pa11y/lib/action.js:29:15)
    at runActionsList (/Users/kkoskelin/projects/pa11y/lib/pa11y.js:315:10)
    at runPa11yTest (/Users/kkoskelin/projects/pa11y/lib/pa11y.js:122:8)
```
which means the call to `page.evaluate( (fn), selector, checked);` isn't getting the expected arguments.  I'm stopping there for the evening.

I'm definitely up for pairing, although I've learned sometimes we need to agree on precisely what that means.  Ping me in slack and we can discuss w/o cluttering the pull request queue. :)

mickyginger:
> Let's take this conversation to the pa11y slack, if that's okay with you? pa11y.slack.com

I can't seem to log in to that workspace @kkoskelin, I think perhaps I need to be added to the workspace in advance.

I think the `evaluate` method in playwright has a different signature, so I can certainly look into that in the meantime.

mickyginger:
Yep that's definitely got me a lot closer, thanks @kkoskelin, I'll have another play this week 👍

mickyginger:
Looks like I'm pretty much done here, except for `test/integration/cli/actions-click-element.test.js`

`runPa11yCli` is timing out when the clicked element does not exists. It's strange because other tests for elements that don't exist seem to behave as expected.

mickyginger:
All tests are now passing (bar the issue on v12.19.1 and warning which I mentioned above).

kkoskelin (approved):
Nice!

paazmaya:
This PR gave me inspiration to replace `puppeteer` with `playwright` in some of my personal projects. Thank you for that @mickyginger

mickyginger:
This branch has been open for a while now. Should I close it, or is there still a chance it'll be shipped?

Zack24441:
Hi,

Could we please fix conflicts and merge these changes? I would really appreciate that change.

Thanks!

Zack24441:
Hi,

Could we please fix conflicts and merge these changes? I would really appreciate that change.

Thanks!

josebolos:
Hi @Zack24441,

As discussed in https://github.com/pa11y/pa11y/issues/534#issuecomment-788030968 at the moment we have no intention of migrating away from puppeteer.

If you require support for Playwright I suggest forking this repo and manually fixing the merge conflicts, not much has changed since this PR was created so it should be pretty straightforward.

I'll close this PR now so there's no confusion around this.

PR #539, opened 2020-11-12
Use the correct SPDX identifier for LGPL

LorenzoAncora:
This software is licensed under LGPL 3.0, without the `or later` legal clause. This pull request ensures that the correct [SPDX](https://spdx.org/licenses/) identifier (`LGPL-3.0-only`) is shown, solving the ambiguity present in the previous one (the SPDX identifier `LGPL-3.0` is deprecated).


TechnologyClassroom:
Why would this project be licensed under `LGPL-3.0-only` instead of `LGPL-3.0-or-later`?  [LGPL-3.0-or-later](https://spdx.org/licenses/LGPL-3.0-or-later.html) is better to avoid potential incompatibility with future versions of the LGPL.

LorenzoAncora:
> Why would this project be licensed under `LGPL-3.0-only` instead of `LGPL-3.0-or-later`? [LGPL-3.0-or-later](https://spdx.org/licenses/LGPL-3.0-or-later.html) is better to avoid potential incompatibility with future versions of the LGPL.

Michael "@TechnologyClassroom", the `LGPL-3.0` SPDX identifier is equivalent to `LGPL-3.0-only` and not to `LGPL-3.0-or-later`. I just corrected the identifier by replacing it with the most recent one. Only the project developers can change or replace the existing license and this pull request does not change the existing license, it only corrects its SPDX identifier.

LorenzoAncora:
sangitamane months have passed, would you be so kind to do the merge?

sangitamane (approved):

PR #546, opened 2021-01-15
Update dependencies and include dependency version information in the package

paazmaya:


paazmaya:
Looks like Travis CI is stuck, would @josebolos, @rowanmanning, or @hollsk have access to see what is happening there?

josebolos:
Hi @paazmaya, it looks like we've run out of Travis credits so unfortunately builds won't run any more until we can migrate the whole project to an alternative.

paazmaya:
@josebolos would you be able to seee if this PR can be merged, perhaps test locally if required?

PR #562, opened 2021-04-07
Extract option functions to a new separate module

josebolos:
Extracts the functions in `lib/pa11y` related to argument parsing and option processing (`parseArguments`, `verifyOptions`, `defaultOptions` and `sanitizeUrl`) to a separate module.

Only `parseArguments` and `verifyOptions` function is used somewhere else, so the module has a clean, simple interface.

No changes to the existing tests were necessary in order for them to pass, but I've extracted the parts of the tests related to these options to their own file.

### DIFF
--- a/lib/option.js
+++ b/lib/option.js
@@ -1,0 +1,100 @@
+'use strict';
+
+const extend = require('node.extend');
+const path = require('path');
+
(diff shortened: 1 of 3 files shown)


hollsk (approved):
👍

PR #563, opened 2021-04-07
Remove reference to PhantomJS from troubleshooting

### LINKED ISSUE #560: Remove references to PhantomJS from troubleshooting guide
The [Troubleshooting guide](https://github.com/pa11y/pa11y/blob/master/TROUBLESHOOTING.md) contains [references to PhantomJS](https://github.com/pa11y/pa11y/blob/master/TROUBLESHOOTING.md#phantomjs-fails-to-run-and-returns-exit-code-127), which we haven't used in a while.

We should remove them in order to avoid confusion.

josebolos:
PhantomJS is not used anymore since v5, so having it there will only confuse people.

Closes #560.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -1,6 +1,4 @@
-
 # Troubleshooting
-
 
 ## Common Issues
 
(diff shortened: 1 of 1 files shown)


hollsk (approved):
👍

Issue #555, opened 2021-03-02
Remove all arrow functions from test suite


josebolos:
Applies to all recent versions of pa11y.

**The problem:** We've had a somewhat unreliable test suite since v5 was release, with some tests occasionally timing out on CI, or even randomly failing.

The move from Travis CI to Github Actions has helped us find at least one test setup that is particularly taxing and was causing the tests to fail, this is being fixed as part of #554.

As part of that issue, it has been necessary to revert some of the changes that made all the test functions in our test suite use arrow functions, as that prevents the individual configuration of those tests using e.g. `this.timeout()`.

[Mocha generally advises against using arrow functions in tests](https://mochajs.org/#arrow-functions):

> Passing arrow functions (aka “lambdas”) to Mocha is discouraged. Lambdas lexically bind this and cannot access the Mocha context.
> [...]
> If you do not need to use Mocha’s context, lambdas should work. Be aware that using lambdas will be more painful to refactor if the need eventually arises!

From _Node.js Web Development - Fifth Edition_ by David Herron:

> With Mocha, it is important to not use arrow functions in the describe and it blocks. [...] Mocha calls these functions with a this object containing useful functions for Mocha. Because arrow functions avoid setting up a this object, Mocha would break.

The Mocha project follows this same pattern on their own tests, e.g.:
https://github.com/mochajs/mocha/blob/master/test/unit/mocha.spec.js

I suggest we remove all the arrow functions from the tests. This will also require changes to the linting config so eslint stops complaining about the lack of arrow functions. This will also make it easier to understand which tests other are flaky and fix the configuration for those tests that are particularly slow.

josebolos:
Fixed in #568.

PR #568, opened 2021-05-17
Replaced arrow functions with normal functions from test suite

sangitamane:
#555 
1. Replaced all arrow functions with normal function style.
2. Updated ESLint config to match the same.


josebolos (approved):
Excellent!

PR #571, opened 2021-05-19
Merge reporter cli

joeyciechanowicz:
Merges the `pa11y-reporter-cli` repo into `pa11y`, preserving the git history of the reporter. 

The only change required to `pa11y` to support this is to first check `../lib/reporters/${name}` for the reporter. The reporter itself has intentionally not been changed at all, however once all reporters are merged we can drop the version bits and bobs and speed up how we load the internal reporters (i.e. keep a map of internal reporters which we can `require` directly without a try/catch)

### DIFF
--- a/bin/pa11y.js
+++ b/bin/pa11y.js
@@ -197,6 +197,7 @@ function loadReporter(name) {
 
 	try {
 		reporterMethods = requireFirst([
+			`../lib/reporters/${name}`,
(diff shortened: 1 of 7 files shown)


joeyciechanowicz:
I based this branch of `main` instead of `master`, going to re-merge on a branch based off of `master` as the rebase was too painful.

PR #583, opened 2021-05-26
Add 6.0 to the migration guide

josebolos:
Also reformats the document so it matches the format suggested by markdownlint and used by other documents in the repo.


sangitamane (approved):
👍

PR #586, opened 2021-06-01
Merge runners

joeyciechanowicz:
Merges both `htmlcs` and `aXe` runners into the repo with their full git histories.

The only substantial change is the removal of using `async` to load the runner scripts and instead opting for `fs.readFileSync` which makes little change given we were loading the runners in a loop and `await`'ing on each read anyway. This results in simplified loading logic and removes a problem when running the tests with coverage 🎉


josebolos (approved):
Looks great! 👍

PR #592, opened 2021-06-25
Adds ignore-rules section to axe

joeyciechanowicz:
Disables rules in axe using the `options.ignore` array. To enable the rules to be cross-compatible with HTML_CodeSniffer we add filtering based on the aXe ruleset.

Fixes https://github.com/pa11y/pa11y-ci/issues/145


lib/runners/htmlcs.js
```diff
@@ -48,13 +48,15 @@ runner.run = async options => {
...
-				}
+		if (!options.rules.length || options.standard === 'Section508') {
+			return;
+		}
+
+		for (const rule of options.rules) {
+			if (window.HTMLCS_WCAG2AAA.sniffs.includes(rule)) {
+				window[`HTMLCS_${options.standard}`].sniffs[0].include.push(rule);
+			} else {
+				throw new Error(`${rule} is not a valid WCAG 2.0 rule`);
```
hollsk, line 59:
```suggestion
				throw new Error(`${rule} is not a valid WCAG 2.1 rule`);
```

lib/runners/htmlcs.js
```diff
@@ -48,13 +48,15 @@ runner.run = async options => {
...
 	 */
 	function configureHtmlCodeSniffer() {
-		if (options.rules.length && options.standard !== 'Section508') {
-			for (const rule of options.rules) {
-				if (window.HTMLCS_WCAG2AAA.sniffs.includes(rule)) {
-					window[`HTMLCS_${options.standard}`].sniffs[0].include.push(rule);
-				} else {
-					throw new Error(`${rule} is not a valid WCAG 2.0 rule`);
-				}
+		if (!options.rules.length || options.standard === 'Section508') {
```
hollsk, line 51: we removed Section 508 as a ruleset option in the 6.0 release. Though can the option can still be passed through to pa11y from Pa11y-CI?

hollsk (commented):
Joey you are a hero. A hero!!

lib/runners/htmlcs.js
joeyciechanowicz: Hmmm I just did this to remove a lint warning. As we filter the rules below based on `HTMLCS_WCAG2AAA` anyway, I think we can drop this.

hollsk (approved):
✨

PR #594, opened 2021-06-26
Cleanup after integration of runners and reporters

### LINKED ISSUE #593: Cleanup after integration of runners and reporters
The following issue were found after integration of the runners and reporters:

- The `pa11y-runner-*` modules are still listed as pa11y `dependencies`
- The test-unit script is excluding the subdirectories to `test/unit/lib`, which includes the reporter and runner folders on linux systems.  The glob pattern is expanded by the shell before passing the results to `mocha`, and it only finds the files in the `tests/unit/lib` folder (see [this issue](https://github.com/mochajs/mocha/issues/1828)).  Single quoting the glob pattern fixes the issue. Note on Windows this does not occur (and tests were failing, see below).
- There are a couple of places in the axe runner unit tests with incorrect paths (see [here](https://github.com/pa11y/pa11y/blob/5fdc60fc42cac29d456460dd87fa5fb6108203ae/test/unit/lib/runners/axe.test.js#L119) and [here](https://github.com/pa11y/pa11y/blob/5fdc60fc42cac29d456460dd87fa5fb6108203ae/test/unit/lib/runners/axe.test.js#L136))
- The reporter unit test were brought over from the existing projects, but aren't named `*.test.js` (just `*.js`), so they're not executed (or wouldn't be even with the fix above for the subfolders)
- The `cli` and `csv` reporter tests have a couple of path issues (see [here](https://github.com/pa11y/pa11y/blob/5fdc60fc42cac29d456460dd87fa5fb6108203ae/test/unit/lib/reporters/cli.js#L13) and [here](https://github.com/pa11y/pa11y/blob/5fdc60fc42cac29d456460dd87fa5fb6108203ae/test/unit/lib/reporters/csv.js#L9))
- The `test-coverage` script has an exclusion for the runners

aarongoldenthal:
This PR includes several changes required after the integration of pa11y runners and reporters into pa11y (fixes #593):

- Removed `pa11y-runner-*` dependencies since now integrated with `pa11y`
- Updated the `test-unit` script to include all subdirectories, which was previously preventing the runner and reporter unit tests from running
- Fixed path issues in axe runner unit tests
- Renamed reporter unit test files so they're executed
- Fixed path issues in the `cli` and `csv` reporter tests
- Removed unit test code coverage exclusion for the runners


joeyciechanowicz (approved):
👏 fantastic catch!

Issue #576, opened 2021-05-24
Improve the logic for detecting remote URLs vs local files


josebolos:
Pa11y is able to test local html files, but it seems to default to remote URLs whenever an argument passed to it is ambiguous:

Ambiguous, fails to run:
```sh
pa11y index.html

Welcome to Pa11y

 > Running Pa11y on URL http://index.html

Error: net::ERR_NAME_NOT_RESOLVED at http://index.html
[...]
```

Unambiguous, runs successfully:
```sh
pa11y ./index.html

Welcome to Pa11y

 > Running Pa11y on URL file:///Users/jose.bolos/index.html

Results for URL: file:///Users/jose.bolos/index.html
[...]
```

We should improve the logic around parsing of arguments so it doesn't fail horribly on local files that do actually exist.

## Environment:

```
Pa11y:      5.3.1
Node.js:    16.1.0
npm:        7.11.2
OS:         20.3.0 (darwin)
```

Pa11y v6 behaves similarly so far.

josebolos:
Fixed in [v6.1.0](https://github.com/pa11y/pa11y/blob/master/CHANGELOG.md#610-2021-11-09) via #598.

PR #597, opened 2021-06-30
576: improve detection of local file paths

kkoskelin:
Addresses #576 
uses `fs.accessSync` to check for existence of a file path on provided url if the url does not begin with "." or "/".
If path exists, the `file://` protocol will be used.  Otherwise it will attempt to use `http://`.


kkoskelin:
messed up the base (2x!)

PR #598, opened 2021-06-30
576: check local paths when parsing options

kkoskelin:
Per #576 attempting to detect valid local paths before deciding to use `http://` scheme

Changed some style and coverage within tests as well. I do not believe that `options` needs to be re-required within the test, and in fact we can import the functions we need directly, just once.

### DIFF
--- a/lib/option.js
+++ b/lib/option.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const extend = require('node.extend');
+const fs = require('fs');
 const path = require('path');
 
 module.exports.parseArguments = parseArguments;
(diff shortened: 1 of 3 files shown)


lib/option.js
```diff
@@ -79,22 +80,35 @@ function defaultOptions(options, defaults) {
...
+}
+
+/**
+ * Check whether a file exists on a given path without regard to permissions
+ * @private
+ * @param {String} filePath - the path to check
+ * @returns {Boolean} true if file exists, false if not
+ */
+function fileExists(filePath) {
+	try {
```
joeyciechanowicz, line 108: You can use `fs.existsSync` and then drop the whole function

joeyciechanowicz (changes requested):
One minor change, the this is 🚀 

Great test additions

lib/option.js
kkoskelin: I've attempted this, but tests fail.  I cannot explain why a call to any of the `fs` functions (existsSync or accessSync) produce errors when they are not captured within a try/catch.
If I call them directly, I get
```
TypeError: fs.existsSync is not a function
```
try rewriting the above function to call any function you desire within a try/catch and it will succeed. Call it directly, and it fails.  I have no explanation, but I'm sure someone here knows something about the *sync functions who could shed light on this.

joeyciechanowicz: It needs to be stubbed out in the mocks first, I think.

https://github.com/pa11y/pa11y/blob/c102611acde655e79043995f5ed73345024340a5/test/unit/mock/fs.mock.js#L6

joeyciechanowicz (approved):
🌟

PR #603, opened 2021-08-11
Replaced WCAG 2.0 mentions with WCAG 2.1, ruleset link updated.

sangitamane:
1. [HTML Code Sniffer WCAG2AAA](https://github.com/pa11y/pa11y/wiki/HTML-CodeSniffer-Rules) ruleset link updated with the [Squizlab's ruleset](https://squizlabs.github.io/HTML_CodeSniffer/Standards/WCAG2/).
2. Will delete the wiki page once this PR is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ Usage: pa11y [options] <url>
     -w, --wait <ms>                the time to wait before running tests in milliseconds
     -d, --debug                    output debug messages
     -S, --screen-capture <path>    a path to save a screen capture of the page to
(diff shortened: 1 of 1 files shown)


allynalford:
Does this mean using WCAG2AA will run WCAG 2.1 rulesets?

josebolos:
Hi @allynalford!

That's right, as WCAG 2.1 doesn't change the existing rules, but instead it only adds new ones, we considered that supporting two separate sets of rules would be too much effort for us, so `WCAG2AA` will use whatever ruleset HTML CodeSniffer implements, which is WCAG 2.1 right now, and hopefully WCAG 2.2 at some point in the future.

PR #613, opened 2021-11-09
Add `clear-field-value` action.

### LINKED ISSUE #612: request: clear field or set field to empty.
## Expected behaviour
For an action we would like to be able to clear a field but the `set field` action in the workflow does not allow for this rule. [The regex](https://github.com/pa11y/pa11y/blob/d602acaad1db8ed90f55ab59cf73ebb3e782c768/lib/action.js#L85) in those code does not allow for empty setting such as undefined or null.

## Actual behaviour

```js
'set field #id to ' // does not run properly as it does not match rule.
'set field #id to ""' //sets the fields value to quotes.
```

## Environment:
```
  System:
    OS: macOS 10.15.7
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 305.06 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.15.5 - ~/.nvm/versions/node/v14.15.5/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.15.5/bin/yarn
    npm: 6.14.11 - ~/.nvm/versions/node/v14.15.5/bin/npm
  pa11y: 6.0.1
```

amprew:
To resolve #612 - a PR to implement an action to allow for clearing fields.

### DIFF
--- a/lib/action.js
+++ b/lib/action.js
@@ -124,7 +124,14 @@ module.exports.actions = [
 					if (!target) {
 						return Promise.reject(new Error('No element found'));
 					}
-					target.value = '';
+					const prototype = Object.getPrototypeOf(target);
+					const {set: prototypeValueSetter} =
+						Object.getOwnPropertyDescriptor(prototype, 'value') || {};
+					if (prototypeValueSetter) {
+						prototypeValueSetter.call(target, '');
+					} else {
+						target.value = '';
+					}
 					target.dispatchEvent(new Event('input', {
 						bubbles: true
 					}));
--- a/lib/action.js
+++ b/lib/action.js
@@ -111,6 +111,31 @@ module.exports.actions = [
 		}
 	},
 
+	// Action to clear an input field value
+	// E.g. "clear field #username"
+	{
+		name: 'clear-field-value',
+		match: /^clear( field)? (.+?)$/i,
+		run: async (browser, page, options, matches) => {
+			const selector = matches[2];
+			try {
+				await page.evaluate(targetSelector => {
+					const target = document.querySelector(targetSelector);
+					if (!target) {
+						return Promise.reject(new Error('No element found'));
+					}
+					target.value = '';
+					target.dispatchEvent(new Event('input', {
+						bubbles: true
+					}));
+					return Promise.resolve();
+				}, selector);
+			} catch (error) {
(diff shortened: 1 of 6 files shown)


lib/action.js
```diff
@@ -111,6 +111,31 @@ module.exports.actions = [
...
+		run: async (browser, page, options, matches) => {
+			const selector = matches[2];
+			try {
+				await page.evaluate(targetSelector => {
+					const target = document.querySelector(targetSelector);
+					if (!target) {
+						return Promise.reject(new Error('No element found'));
+					}
+					target.value = '';
+					target.dispatchEvent(new Event('input', {
```
amprew, line 128: thinking between `input` and `change` events.

amprew:
If accepted I will add docs also to README and other markdown files.

lib/action.js
joeyciechanowicz: You do want the `input` event. If you look at the [set-field-value](https://github.com/pa11y/pa11y/blob/1d31213b7403c09456a1f2f38aebef0182fd4552/lib/action.js#L95) action theres some extra bit that need to be performed to trigger any `setters` that have been added (such as in Reacts case).
The common functionality of these two actions could definitely be abstracted

amprew: Thanks for the comment. I have added in this now

joeyciechanowicz (approved):
This is wonderfully thorough ⭐  I especially liked the test that removes the image, real nice!

amprew:
Thanks, @joeyciechanowicz! Do you know what the process is merging from here? Do I need to create a changelog?

joeyciechanowicz:
> Thanks, @joeyciechanowicz! Do you know what the process is merging from here? Do I need to create a changelog?

Nothing from you, I'll merge it on Monday when I can do a release as well (so the docs and package are in sync)

joeyciechanowicz:
GitHub has done something strange. I've squashed and merged this PR, and it is now on master, however GH seems to think that it isn't. I'm going to close the PR now, rather than work out why GH broke. Thanks for adding this!!

amprew:
Thanks! Glad it's in.

PR #618, opened 2021-12-02
Update README.md

gabalafou:
htmlReporter.results(results) not results(results, url)


gabalafou:
See https://github.com/pa11y/pa11y/blob/master/lib/reporters/html.js#L15

joeyciechanowicz (approved):
Good catch

PR #625, opened 2022-03-02
chore(lib): fix app runtime required modules

j-mendez:
1. fix required runtime modules for library. Commander is only for the cli via bin.


josebolos:
Hi @j-mendez,

Thanks for your PR. `commander` is a runtime dependency, it's necessary for the users to be able to [run pa11y from the command line](https://github.com/pa11y/pa11y#command-line-interface), so it needs to be a `dependency`. Dev-dependencies will not be installed when running for example `npm install --global pa11y`.

PR #626, opened 2022-03-08
Fix README shield build badge

manuelmeister:


josebolos (approved):
Good catch! Thanks a lot for the PR, @manuelmeister!

PR #628, opened 2022-03-10
Replace unmaintained `hogan.js` (used by the html reporter) dependency with maintained `mustache.js` 

wit-0-bit:
### Rationale:
While using pa11y from ClojureScript, I encountered an error when the ClojureScript compiler tried to import the node module `hogan.js`. ClojureScript evidently gets confused about whether or not it's a javascript file named `hogan.js` or a module named `hogan.js`.

Regardless, while investigating that, I noticed that Hogan is no longer maintained: https://github.com/twitter/hogan.js/commit/c91c0234e0f7418d27d5ccb4698e54360f15809b

So, I figured I would make a PR replacing `hogan.js` with `mustache.js` https://github.com/janl/mustache.js, which should both solve my admittedly very niche and likely unsupported issue and move pa11y off of a no-longer-maintained dependency. 

That said, I understand if that isn't sufficient reason to swap out the templating library for the HTML Reporter.

### Features
In theory, this should be a simple drop-in replacement as they both implement the mustache syntax.  Despite that, I did have to use `&` in front of the page url in the template due to some automatic url encoding that `mustache.js` did.


wit-0-bit:
Hi @josebolos,

Thank you for the update! This is currently non-blocking for us (we are using 6.0.1 in the interim, which is before the HTML reporter (and therefore Hogan) was merged into this repo) so no worries if it is a while before you can take a look 😄 

Thanks again!

josebolos (approved):
This looks great to me! Thank you so much for the PR!

PR #640, opened 2022-04-08
Add windows and macos to the test matrix

josebolos:

PR #644, opened 2022-04-09
Fix Windows specific issues

aarongoldenthal:
This is an update to https://github.com/pa11y/pa11y/pull/640 to fix the Windows-specific issues (based heavily on those identified in https://github.com/pa11y/pa11y-ci/pull/177). Note this merges into the `github-actions-all-os` branch in https://github.com/pa11y/pa11y/pull/640, not `master`.

- Set default environment variable value within global-setup (to remove platform-specific commands in `package.json` to set the default, but still allow another to be set)
- Update handling of absolute paths in Windows (in `options.js`)
  - Uses `path.isAbsolute` for better handling between *nix and Windows. Using the wrong absolute path format for the OS produces erroneous results (e.g. running `pa11y /foo/bar.html` in Windows, or `pa11y E:\some\folder\index.html` in *nix), but the cases are broken before an after this change, and that's not necessarily unexpected.
  - Also note this is different than results from the `protocolify` module, which was wrong in some cases, and could produce some inconsistent pa11y-ci results (since `protocolify` is use for URLs specified via the CLI and the rest (sitemap or config) are passed directly to pa11y.
- Fix Windows absolute path issues in `lib/pa11y.js`, several tests, and `createMockWebsite`
- Update `runPa11yCli` to call the `node` command explicitly since Windows ignores the shebang in `bin` files
- Increased test timeout to 20s since some integration test were failing on Mac with 10s (not sure this has had enough runs to say for sure that fixed it, it's inconsistent).

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -13,6 +13,5 @@
 	},
 	"globalSetup": "./test/integration/global-setup.js",
 	"globalTeardown": "./test/integration/global-teardown.js",
-	"setupFiles": ["./test/setup.js"],
 	"setupFilesAfterEnv": ["./test/integration/setup-env.js"]
 }
--- a/jest.config.json
+++ b/jest.config.json
@@ -13,5 +13,6 @@
 	},
 	"globalSetup": "./test/integration/global-setup.js",
 	"globalTeardown": "./test/integration/global-teardown.js",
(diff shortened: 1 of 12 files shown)


josebolos (approved):
Amazing! Thanks so much for this! 🎉

Issue #662, opened 2022-11-09
TimeoutError: Pa11y timed out 


davidwilby:
## Expected behaviour
run `pa11y http://localhost:4000/` for locally served website, get output

## Actual behaviour
```
pa11y -d http://localhost:4000

Welcome to Pa11y

 > Running Pa11y on URL http://localhost:4000
 > Debug: Launching Headless Chrome

Error: TimeoutError: Pa11y timed out (6000ms)
    at Timeout._onTimeout (/home/<user>/miniconda3/envs/<environment>/lib/node_modules/pa11y/node_modules/p-timeout/index.js:39:64)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7)
```

Note: pa11y is installed with npm in a conda environment

I've tried:
- increasing the timeout
- double-checking the locally running web server
- running against public websites
- Including `{ "chromeLaunchConfig": { "ignoreHTTPSErrors": false } }` in `./pa11y.json`
- an `nvm` install of node

All with the same results

## Steps to reproduce
1. Install nodejs in conda environment on ubuntu 22.04
2. `npm install pa11y`
3. `pa11y https://localhost:4000`

## Environment

```
  System:
    OS: Linux 5.17 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 2.36 GB / 15.27 GB
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.11.0 - ~/miniconda3/envs/<environment>/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.19.2 - ~/miniconda3/envs/<environment>bin/npm
  pa11y: 6.2.3
```

davidwilby:
Seems like it might be an ubuntu 22.04 issue, see #651

jankonzack:
Tried successfull running pa11y with a config setting "executablePath" within "chromeLaunchConfig" using Chromium (107.0.5304.87 snap).

`{
    "chromeLaunchConfig":
    {
        "ignoreHTTPSErrors": false,
        "executablePath": "/snap/bin/chromium"
    }
}`

`System:
    OS: Linux 5.19 Ubuntu 22.10 22.10 (Kinetic Kudu)
[...]

Issue #651, opened 2022-07-10
pa11y fails with Github Actions ubuntu-22.04


bansan85:
I tried to update CI Github Actions of my project from 'ubuntu-latest' to 'ubuntu-22.04' (beta).

Now, pa11y always runs on timeout.

I tried to update CI Github Actions of pa11y project. As you can see, running tests fails (https://github.com/bansan85/pa11y/runs/7273637127)

davidwilby:
Similar problem on ubuntu 22.04 desktop, reported in #662

hollandjg:
GitHub is currently switching over from ubuntu 20.04 to 22.04 as the `ubuntu-latest` runner, so this issue will become much more prevalent. We had to deal with it too.
The ability to get the debug messages from pa11y (or the log files) would have been super useful when debugging this. The way we worked out what was happening in the end was to add an additional step in the GitHub actions which ran `pa11y` with the `--debug` flag and that showed that ChromeHeadless just wasn't loading.

PR #664, opened 2022-12-03
Update dependencies to resolve Ubuntu issues

aarongoldenthal:
Update dependencies to latest versions, including Puppeteer, to resolve issues with Ubuntu 22.04, which is [now `ubuntu-latest` for GitHub actions](https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/). Closes https://github.com/pa11y/pa11y/issues/662 and https://github.com/pa11y/pa11y/issues/651.

- Update tests to latest Node releases - BREAKING
  - Drop v12 (EOL 2022-04-30), v14 (EOL 2023-04-30), v16 (EOL 2023-09-11)
  - Skip v19 (EOL 2023-06-01)
  - Add v18 and v20
- Update to lockfile v2
- Update to latest devDependencies (`eslint`, `eslint-plugin-jest`, `jest`)
- Update dependencies with patch updates (`kleur`)
- Update dependencies with minor updates (`axe-core`, `semver`)
- Update dependencies with major updates (`puppeteer`, `commander`)
  - This includes 10 major version updates to `puppeteer` (9.1.1 to 19.10.1), and is a BREAKING change (especially the new binary install script in v19 and Chrome updates from 92 to 112)

Note, did not update `p-timeout` since v5 is pure ESM.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
             lint: true # Linter is run only once to shorten the total build time
           - node-version: 16.x
           - node-version: 18.x
-          - node-version: 19.2.x
+          - node-version: 19.x
 
     steps:
       - name: Checkout code from ${{ github.repository }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,6 +28,7 @@ jobs:
           - node-version: 16.x
           - node-version: 18.x
           - node-version: 19.x
+          - node-version: 20.x
 
     steps:
       - name: Checkout code from ${{ github.repository }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,10 +24,8 @@ jobs:
       matrix:
         include:
           - node-version: 16.x
-            lint: true # Linter is run only once to shorten the total build time
           - node-version: 18.x
           - node-version: 20.x
-
     steps:
       - name: Checkout code from ${{ github.repository }}
         uses: actions/checkout@v2
@@ -38,7 +36,6 @@ jobs:
       - name: Install dependencies
         run: npm ci
       - name: Run linter
-        if: ${{ matrix.lint }}
         run: npm run lint
       - name: Run tests
         run: npm test
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
             lint: true # Linter is run only once to shorten the total build time
           - node-version: 16.x
           - node-version: 18.x
-          - node-version: 19.x
+          - node-version: 19.2.x
 
     steps:
       - name: Checkout code from ${{ github.repository }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,11 +23,9 @@ jobs:
     strategy:
       matrix:
         include:
-          - node-version: 14.x
-            lint: true # Linter is run only once to shorten the total build time
           - node-version: 16.x
+            lint: true # Linter is run only once to shorten the total build time
           - node-version: 18.x
-          - node-version: 19.x
           - node-version: 20.x
 
     steps:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,10 +23,10 @@ jobs:
     strategy:
       matrix:
         include:
-          - node-version: 12.x
-            lint: true # Linter is run only once to shorten the total build time
           - node-version: 14.x
+            lint: true # Linter is run only once to shorten the total build time
           - node-version: 16.x
+          - node-version: 18.x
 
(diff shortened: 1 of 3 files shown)


aarongoldenthal:
The GitHub Actions are pinned to `19.2.x` since tests are failing in 19.3.0 when trying to assign to `gloval.Event` as follows (the mock before the test, and original value after the test):

```sh
  ● lib/action › is a function

    TypeError: Cannot assign to read only property 'Event' of object '[object global]'

      16 |              mockEvent = {event: true};
      17 |              originalEvent = global.Event;
    > 18 |              global.Event = jest.fn().mockReturnValue(mockEvent);
         |                          ^
      19 |
      20 |              mockElement = createMockElement();
      21 |              originalDocument = global.document;

      at Object.<anonymous> (test/unit/lib/action.test.js:18:15)

  ● lib/action › is a function

    TypeError: Cannot assign to read only property 'Event' of object '[object global]'

      26 |
      27 |      afterEach(() => {
    > 28 |              global.Event = originalEvent;
         |                          ^
      29 |              global.document = originalDocument;
      30 |      });
      31 |

      at Object.<anonymous> (test/unit/lib/action.test.js:28:15)
```

aarongoldenthal:
The test issues above were resolved in a subsequent version of Node 19.

meekay:
Hi, are there plans to merge this in soon? The current version of Puppeteer is old enough that it doesn't seem to support Apple silicon.

danyalaytekin:
Hi @aarongoldenthal.  Thanks for your work on this.  I'm taking a look, first of all to find out how this runs a few months on, since the previous logs have expired.  I've reverted a couple of my commits there rather than resetting straight back to https://github.com/pa11y/pa11y/pull/664/commits/9d74f5d332e68836742020a023f6fce2eb7f218f, since you probably have this branch checked out and I didn't feel comfortable force pushing; sorry for the extra noise.

aarongoldenthal:
@danyalaytekin If this is getting reviewed I'm happy to make another update to the latest, I just held off since it wasn't, but at least a few of these dependencies have fallen behind.

danyalaytekin:
@aarongoldenthal Thank you, potentially, but no need to spend that time just yet; I'm still getting up to speed here.

danyalaytekin:
Thanks again @aarongoldenthal.  I've pushed another change to drop Node 16 in addition to 12 and 14 since it reached EOL between the time of your PR being submitted and reviewed.  I did also force-push about an hour ago to remove the noisy commits I made while trying to bump CI last week.  Hope this didn't cause you any trouble locally.

### Note for reviewers

While some of these dependencies have been versioned further since the PR was submitted, we can bump those later on and I'd like us to merge this PR first to fix the main problem.

Re: branching and `next`. This PR's target is the default branch.  https://github.com/pa11y/pa11y/pull/647 has already arrived in `next` but, bearing in mind I might be unaware of some past pain, I agree with `master` ([soon](https://github.com/pa11y/org/issues/19) to become `main`) being the target for this PR and in general for anything which isn't a patch fix that will never arrive in the default branch.  If you agree with this approach please let me know and I'll document it in some more detail, and will also make a PR to merge `next` into the default branch; there will probably be a couple of merge issues but nothing too serious.

I'm comfortable with the other major updates:

#### Commander 

[Commander's breaking changes](https://github.com/tj/commander.js/blob/master/CHANGELOG.md) are either to Node.js support or look benign.

#### ESLint

I've checked [over the migration guide for 7 → 8](https://eslint.org/docs/latest/use/migrate-to-8.0.0) and it looks ok.  I introduced some errors locally to `lib` and then `test` and linting for each failed reassuringly. 

#### Jest

We don't use snapshots or `jest.mocked` which are among the few items which saw [breaking changes in Jest 29](https://jestjs.io/docs/upgrading-to-jest29). There were many more changes in Jest 28 and [the list is long](https://github.com/jestjs/jest/blob/main/CHANGELOG.md#2800). So I'm relying partially on the test suite there, which continues to pass in CI, but I also couldn't spot any obvious problems in [28's own migration guide](https://jest-archive-august-2023.netlify.app/docs/28.x/upgrading-to-jest28).

.github/workflows/tests.yml
```diff
@@ -22,12 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        include:
-          - node-version: 12.x
-            lint: true # Linter is run only once to shorten the total build time
```
danyalaytekin, line 27: Reasoning for removing this logic:

👎 Will now take 2-5s longer per job

👍 In comparison, running tests takes over 3m per job
👍 Slightly increase confidence
👍👍👍 Reduce complexity

jpw (approved):
LGTM, thanks!

danyalaytekin:
Thanks both and @aarongoldenthal!

PR #683, opened 2023-11-12
Clean up nomenclature to properly reference runner, not reporter

aarongoldenthal:
It appears some of the reporter loading/validation logic was copied to support runners, but some "reporter" references were overlooked. This cleans up the runner loading/validation logic and tests to properly reference `runner` instead of `reporter`.

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -378,15 +378,15 @@ function loadRunnerFile(runner) {
  * @param {String} runnerName - The name of the runner.
  * @param {String} runnerSupportString - The runner support string (a semver range).
  * @param {String} pa11yVersion - The version of Pa11y to test support for.
- * @throws {Error} Throws an error if the reporter does not support the given version of Pa11y
+ * @throws {Error} Throws an error if the runner does not support the given version of Pa11y
  * @returns {void}
  */
-function assertReporterCompatibility(runnerName, runnerSupportString, pa11yVersion) {
+function assertRunnerCompatibility(runnerName, runnerSupportString, pa11yVersion) {
 	if (!runnerSupportString || !semver.satisfies(pa11yVersion, runnerSupportString)) {
 		throw new Error([
 			`The installed "${runnerName}" runner does not support Pa11y ${pa11yVersion}`,
 			'Please update your version of Pa11y or the runner',
-			`Reporter Support: ${runnerSupportString}`,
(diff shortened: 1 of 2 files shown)


aarongoldenthal:
@danyalaytekin This test seems to be intermittently failing on macOS, don't think it's specific to this PR (first run was on Node 18, this time on Node 20, also see on other actions runs).

danyalaytekin:
Hey @aarongoldenthal, thank you for this PR and I'll have a look.

Regarding the flake, yes, I added [this extra assertion](https://github.com/pa11y/pa11y/pull/680/files#diff-74c0e9e8494d1b29c270ab8797b6d3e2e5fca5031c548b091a5e4cadbbf7b03bR26) previously to obtain more clarity while trying to diagnose it, increased the timeout and reverted Puppeteer to 20 from 21, which I thought had helped, but it looks like it's still hanging around.  For now I'll run the failing test again.  [This test's fixture](https://github.com/pa11y/pa11y/blob/3fcb2787027cd4ff57265d408b4005a302226f6f/test/integration/mock/html/actions-wait-for-element-event.html) is unusual and something about it is confusing Puppeteer.

danyalaytekin (approved):
Perfect, thanks

danyalaytekin:
Hi again @aarongoldenthal.  Ok, I do think it's a performance issue which is either exacerbated here with GitHub's macOS runner, or which is occurring unpredictably because Jest shuffles the test run order.  I reproduced it locally on some runs by adding `slowMo: 250` to `chromeLaunchConfig` for the test run.

aarongoldenthal:
@danyalaytekin It's interesting that specific test is failing. Looking at some other results, it's not specifically running slow when it passes. It may be coincidence, but in the cases it passes it's in the late 20s, but when it fails it's in the late 30s.

I had speculated a performance issue as you noted, but here I'm just Windows and Linux and didn't have a good platform to play around with.

danyalaytekin:
@aarongoldenthal Thanks for following up. I've created a PR that'll hopefully fix this:

- #684

> when it fails it's in the late 30s

That's an interesting observation and would add confidence to the possibility that the event was somehow being fired, on some runs, before the listener was up and running.

danyalaytekin:
Passing after rebasing.  Thanks again @aarongoldenthal

PR #684, opened 2023-11-13
Fix flake affecting `wait-for-element-event` integration tests

danyalaytekin:
This PR:

1. resolves (hopefully resolves) the test flake that has frequently been causing this test to fail on `macos-latest` in GitHub Actions.  Hypothesis: Puppeteer was dispatching the event before having its listener in place.
1. increases the `--wait` test's legroom from 100ms to 500ms since that has failed occasionally too 
1. reinstates the `notahost` test, since it works in GitHub Actions

Thanks for the nudge @aarongoldenthal

### DIFF
--- a/test/integration/cli/exit-codes.test.js
+++ b/test/integration/cli/exit-codes.test.js
@@ -29,10 +29,9 @@ describe('CLI exit codes', () => {
 
 	});
 
-	// This has to be skipped for now, some ISPs hijack hostnames that can't
-	// be resolved (looking at you TalkTalk). We could do with finding a better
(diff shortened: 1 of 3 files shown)


danyalaytekin:
Heh, triggering the event after 100ms instead of 2s has the [whole group flailing](https://github.com/pa11y/pa11y/pull/684/checks?sha=a6a047e0efa98d454d6e7879083f4bd8fc08d07a).  Now to go the other way...

sangitamane (approved):
⏲️ LGTM

PR #685, opened 2023-11-17
[WIP] Fix axe level translation

### LINKED ISSUE #623: pa11y reports an axe warning as an error
**Please note:** This repository's issues are reserved for feature requests, bug reports, and other issues with the Pa11y library itself.

If you need help integrating Pa11y with your application, we recommend using [Stack Overflow](https://stackoverflow.com/questions/tagged/pa11y).

For a bug report, please use the template below:

---

## Expected behaviour

When an axe rule can't determine something, it should be reported as a warning, not an error. If that sounds vague, it's because I'm trying to extrapolate from the rule on which I see this happening to generalize about any rule in which axe reports _`x` could not be determined due to `y`_.

The specific rule in this case is [color-contrast](https://dequeuniversity.com/rules/axe/4.3/color-contrast). You'll see toward the bottom of that definition page it says:

> This rule will not report on text elements that have a background-image, are obscured by other elements or are images of text.

For instance, if an element has a gradient background (which is a type of background-image in CSS) overlaid with text, looking at it using the axe browser extension it will say:

> Element's background color could not be determined due to a background gradient.

When pa11y receives a notice like this, it should consider it a warning, not an error.

## Actual behaviour

Instances like this are reported as errors:

```
 • Error: Elements must have sufficient color contrast (https://dequeuniversity.com/rules/axe/4.4/color-contrast?application=axeAPI)
```

## Steps to reproduce

Unfortunately running pa11y (or axe-core/cli) against a CodePen doesn't really work, but here's some really simple code to test:
https://codepen.io/rootwork/pen/zYPwQbv

If you inspect that using axe's browser extension, you'll see the warning noted above.

In my case I was running pa11y locally using `pa11y --runner axe http://localhost:9050/` against a page that had similar a text-on-gradient section.

Finally, if I'm wrong about diagnosing this issue and it's actually a problem with axe-core/cli, please accept my apologies and I'll re-report it over there.

## Environment:
```
  System:
    OS: Linux 5.4 Linux Mint 20.3 (Una)
    CPU: (4) x64 Intel(R) Core(TM) i5-6600 CPU @ 3.30GHz
    Memory: 1.13 GB / 23.35 GB
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.13.2 - ~/n/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 8.1.2 - ~/n/bin/npm
  pa11y: 6.1.1
```

danyalaytekin:
Full description pending.

- intended to fix #623 

#### Thoughts

- [ ] **Semver:** I've pencilled this in for `7.0.1`, but would this fix be considered to require a major bump?  So far I've seen a couple of workarounds that ignore rules that are emitting these false positives but I suppose people could be forming a contract with Pa11y's current output.  On the other hand, it isn't behaviour we intended - it's a bug.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "6.2.3",
       "license": "LGPL-3.0-only",
       "dependencies": {
-        "axe-core": "~4.8.2",
+        "axe-core": "^4.8.2",
         "bfj": "~8.0.0",
         "commander": "~11.1.0",
         "envinfo": "~7.11.0",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "npm": ">=10"
   },
   "dependencies": {
-    "axe-core": "~4.8.2",
+    "axe-core": "^4.8.2",
     "bfj": "~8.0.0",
     "commander": "~11.1.0",
     "envinfo": "~7.11.0",


jackfoust:
If I can help test, please let me know. Would love to see this issue with the contrast errors resolved.

danyalaytekin:
Hi @jackfoust and @janus-reith, I hope to return to this shortly after refamiliarising myself.  Time flies.

lib/pa11y.js
```diff
@@ -305,27 +319,33 @@ async function injectRunners(options, state) {
...
+		pa11yVersion,
+		hideElements,
+		ignore,
+		rootElement,
+		rules,
+		runners,
+		standard,
+		wait
+	}
+);
```
aarongoldenthal: It looks like `` is not being passed to the runner:

```suggestion
const runPa11yWithOptions = ({
	hideElements,
	ignore,
	levelCapForReviewRequired,
	rootElement,
	rules,
	runners,
	standard,
	wait
}, state) => state.page.evaluate(
	/* eslint-disable-next-line no-underscore-dangle */
	runOptions => window.__pa11y.run(runOptions),
	{
		pa11yVersion,
		hideElements,
		ignore,
		levelCapForReviewRequired,
		rootElement,
		rules,
		runners,
		standard,
		wait
	}
);
```

With this, the JSON results looked correct, but the CLI reporter was showing `undefined` for the severity and I didn't have a chance to fully sort that out.

test/integration/cli/runner-axe.test.js
```diff
@@ -3,24 +3,17 @@
...
 
-// Note: we use the JSON reporter in here to make it easier
-// to inspect the output issues. The regular CLI output is
-// tested in the reporter tests
-describe('CLI runner aXe', function() {
-	let pa11yResponse;
-
-	before(async function() {
-		pa11yResponse = await runPa11yCli(`${global.mockWebsiteAddress}/errors`, {
+describe('CLI runner axe', function() {
```
aarongoldenthal: It seems like we need some additional integration tests checking for the severity change. This is a test page I created (thanks Copilot) with one issue of each axe type:

```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="UTF-8" />
		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
		<title>Axe Test Page - Multiple Severity Levels</title>
	</head>
	<body>
		<main>
			<h1>Axe Accessibility Test Page</h1>

			<!-- CRITICAL VIOLATION: image-alt - Images must have alternative text -->
			<img src="data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7" width="100" height="100" />

			<!-- SERIOUS INCOMPLETE: color-contrast - Elements must meet minimum color contrast ratio thresholds -->
			<div style="background-image: url('data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7');padding: 20px;">
				<p>Text over a background image - axe cannot automatically determine contrast</p>
			</div>

			<!-- SERIOUS VIOLATION: frame-title - Frames must have an accessible name -->
			<!-- CRITICAL INCOMPLETE: frame-tested - Frames should be tested with axe-core -->
			<iframe src="about:blank" width="300" height="200"></iframe>

			<!-- MODERATE VIOLATION: heading-order - Heading levels should only increase by one -->
			<h2>Heading Structure</h2>
			<h4>This H4 skips H3 level</h4>

			<!-- MINOR VIOLATION: table-duplicate-name - Tables should not have the same summary and caption -->
			<table summary="Sales Data">
				<caption>Sales Data</caption>
				<tr>
					<th>Month</th>
					<th>Sales</th>
				</tr>
				<tr>
					<td>January</td>
					<td>$1000</td>
				</tr>
			</table>
		</main>
	</body>
</html>
```

lib/pa11y.js
danyalaytekin: Yikes, thanks @aarongoldenthal.

lib/runners/axe.js
```diff
@@ -3,215 +3,216 @@
...
+				element,
+				code: id,
+				type: choosePa11yLevel(impact, issueNeedsFurtherReview, levelCapForReviewRequired),
+				message: `${help} (${helpUrl})`,
+				runnerExtras: {
+					description,
+					impact,
+					help,
+					helpUrl
+				}
```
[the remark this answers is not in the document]
danyalaytekin: Thanks, added

lib/pa11y.js
danyalaytekin: Hi again @aarongoldenthal 

1. We could add a [set of choices](https://www.npmjs.com/package/commander#more-configuration) to the input as validation, but this Commander interface (`addOption`) isn't one we're using yet, so ~I'd prefer to do that as part of a pass over the whole Commander interface in case it has changed in the meantime. We're on version 11 and they're up to version 14 (memory fail - we're on 14 too)~ I'll have a look.
2. My ideas:
    - `level-cap-[if, when]-needs-review`
    - `level-cap-[if, when]-incomplete`
    - 🤷🏽‍♂️

aarongoldenthal: Hi @danyalaytekin 
1. I could see going either way. I don't believe we do the same for other inputs, for example `level`, so maybe a separate change for broader input validation.
2. I prefer `when` versus `if`, and I think `needs-review` gives an explicit action, so I'd vote for `level-cap-when-needs-review`

danyalaytekin:
If people are reluctant to review this, I can recreate it as a more minimal PR, and leave the refactoring to another PR (to be approved or otherwise).

aarongoldenthal:
> If people are reluctant to review this, I can recreate it as a more minimal PR, and leave the refactoring to another PR (to be approved or otherwise).

Sorry @danyalaytekin, I just hadn't had time to take another look. This is great, and it looks like my previous comments have been covered, but in running a few scenarios I found one potential issue - `levelCapWhenNeedsReview` is ignored in a config file. Specifically, config is ignored below because the CLI argument has a default value (which we don't have in most places, and thinking about it will then only work via the CLI), so `programOptions.levelCapWhenNeedsReview` always wins
https://github.com/pa11y/pa11y/blob/3a5387ff0f9d72274103b4af2c6c784fd21136d5/bin/pa11y.js#L160-L185

I see a couple of options that worked in my testing (I lean a little towards #2, although I think either would work):

1. Remove the CLI default and add and explicit `levelCapWhenNeedsReview: 'error'` here:
https://github.com/pa11y/pa11y/blob/3a5387ff0f9d72274103b4af2c6c784fd21136d5/bin/pa11y.js#L164-L166

2. Simply remove the CLI default value. In this case the check below will revert to the default `level` from `axeImpactToPa11yLevel` with `levelCapWhenNeedsReview` undefined (I think this works, although did not do exhaustive testing)
https://github.com/pa11y/pa11y/blob/3a5387ff0f9d72274103b4af2c6c784fd21136d5/lib/runners/axe.js#L134-L143

With that, I propose also adding the following to the README for completeness, which I missed before:

```md
### `levelCapWhenNeedsReview` (string)

The level to cap any issue requiring manual review to. This should be one of `error` (the default), `warning`, or `notice`. Only used by the axe runner.

```json
{
    "levelCapWhenNeedsReview": "warning"
}
```
```

File: lib/pa11y.js [hunk 4]
```diff
@@ -4,34 +4,52 @@ const runAction = require('./action');
...
 /**
  * Run accessibility tests on a web page.
+ * Note: actual signature is variadic and parameters shift left on sibling absence.
  * @public
- * @param {String} url - The URL to run tests against.
- * @param {Object} [options={}] - Options to change the way tests run.
- * @param {Function} [callback] - An optional callback to use instead of promises.
- * @returns {Promise} Returns a promise which resolves with a results object.
+ * @param {string} url The URL to run tests against.
+ * @param {object} [options={}] Options to change the way tests run.
```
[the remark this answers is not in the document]
josebolos: > JSDoc's docs use [lower-case for primitives and capitalisation for other things](https://jsdoc.app/tags-type).

Oh that's even more confusing 🙈

PR #694, opened 2024-03-09
Use native `AbortSignal` in place of `p-timeout`

danyalaytekin:
## Summary

Replace the use of [p-timeout](https://github.com/sindresorhus/p-timeout) with Node's native [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static) (Node 15+), given we can no longer update `p-timeout` beyond version 4 due to [the author's stance](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) on ESM and CommonJS.  Parity is maintained with the previous implementation.

## Detail

1. Remove dependency `p-timeout`.
2. In this PR's revision the recipient of the `AbortSignal` does not interact with or listen to it, but this implementation still achieves parity with the previous version.
3. Remove the unit test of the calls to the dependency, without replacing it.  This test was closely tied to the implementation and the behaviour is covered by the corresponding integration test.

## To review timeout presentation locally

```shell
nvm install 20
npx \
    pa11y/pa11y#update-ptimeout-to-abortsignal \
    https://example.com \
    --timeout 1
```

Expected:

```console
Welcome to Pa11y

 > Running Pa11y on URL https://example.com

Error: TimeoutError: Navigation timeout of 1 ms exceeded
    (etc)
```

### DIFF
--- a/test/unit/lib/pa11y.test.js
+++ b/test/unit/lib/pa11y.test.js
@@ -21,7 +21,6 @@ describe('lib/pa11y', () => {
 	let puppeteer;
 	let extend;
 	let semver;
-	let promiseTimeout;
 
 	beforeEach(() => {
 		pa11yResults = {
@@ -56,15 +55,10 @@ describe('lib/pa11y', () => {
 				return jest.fn(actual);
 			});
 			jest.doMock('semver', () => require('../mocks/semver.mock'));
-			jest.doMock('p-timeout', () => {
-				const actual = jest.requireActual('p-timeout');
-				return jest.fn(actual);
-			});
 
 			extend = require('node.extend');
 			puppeteer = require('puppeteer');
 			semver = require('semver');
-			promiseTimeout = require('p-timeout');
 
 			puppeteer.mockPage.evaluate.mockResolvedValue(pa11yResults);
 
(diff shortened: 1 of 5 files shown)


test/unit/lib/pa11y.test.js
```diff
@@ -21,7 +21,6 @@ describe('lib/pa11y', () => {
 	let puppeteer;
 	let extend;
 	let semver;
-	let promiseTimeout;
```
hollsk, line 24: 🚨 Potentially silly question incoming klaxon: why don't we need a unit test on the timeout anymore?

danyalaytekin: I'm not fully convinced one way or the other either.  Overall though, the relevant sections of this file were:

- getting rid of the actual timeout functionality, maybe so it didn't trip up the unit tests
- while making sure that its provider `p-timeout` was still called, somewhere/anywhere during its execution (it isn't checking that a timeout happens or when)

The effects of this PR's changes:

- the default timeout is still in play, but it doesn't occur during the unit tests because they're speedy
- the test of the timeout's actual effect is made by the integration test
    - ... but, as described above, this was the case before
- the fact that the real object rather than a mock is in use could be argued to build confidence in the test (I'm reaching here)

Confidence level: 75%

hollsk: 75% is enough convincing for me! thanks for the explanation :D I am a big newb with this sort of thing so just needed a bit of handholding, heh.

hollsk (approved):
Let's goooo! 🎉

PR #701, opened 2024-03-30
Clean up several minor issues in setPage

aarongoldenthal:
The following issues were noticed today in the `setPage` function (all are low priority cleanup).

- Removed unused `url` argument from `setPage`
- Moved debug log entry for navigating to URL to proper location (from `setPage` to `gotoUrl`)
- Update debug logging for all `setPage` operations

Debug log changes:

```diff
 > pa11y -d https://pa11y.org

 Welcome to Pa11y

  > Running Pa11y on URL https://pa11y.org
  > Debug: Launching Headless Chrome
- > Debug: Opening URL in Headless Chrome
+ > Debug: Opening new browser page
+ > Debug: Setting page User-Agent to "pa11y/8.0.0"
+ > Debug: Setting page viewport
+ > Debug: Navigating to https://pa11y.org
  > Debug: Loading runner: htmlcs
  > Debug: Injecting Pa11y
  > Debug: Injecting runner: htmlcs
  > Debug: Running Pa11y on the page
  > Debug: Document title: "Pa11y"

 No issues found!
```

### DIFF
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -245,7 +245,7 @@ async function interceptRequests(options, state) {
 async function gotoUrl(url, options, state) {
 	// Navigate to the URL we're going to test
 	if (!options.ignoreUrl) {
-		options.log.debug(`Opening URL ${url}`);
+		options.log.debug(`Navigating to ${url}`);
 		await state.page.goto(url, {
 			waitUntil: 'networkidle2',
 			timeout: options.timeout
--- a/lib/pa11y.js
+++ b/lib/pa11y.js
@@ -162,7 +162,6 @@ async function setPage(options, state) {
 	};
 
 	state.page.on('console', state.consoleCallback);
-	options.log.debug('Opening URL in Headless Chrome');
 	if (options.userAgent) {
 		await state.page.setUserAgent(options.userAgent);
 	}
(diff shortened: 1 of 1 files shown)


danyalaytekin (commented):
(Oops, let me try again using the review button.)

Well spotted @aarongoldenthal!

1. I think you've completed a [reversion which hadn't quite concluded](https://github.com/pa11y/pa11y/commit/52af3fc5e6963c26f575ef104c3b36d5ca8c19fa).  But this happened a while ago so I should check... are you happy with this signature change @josebolos or was it left there pending a restoration of the reverted feature?
2. Checking that commit I also noticed that we have some testing around the debug logging, which it could make sense to expand to cover the changes in this PR, for consistency, if that's our practice.  Alternatively the extra logging could be pushed into a separate PR.  @aarongoldenthal @josebolos

josebolos:
Hi all! Sorry I'm a bit late to the party. This seems absolutely fine to me, I tried to find the logs for the GH actions workflows but it looks like the PR is so old that GH has deleted the logs and I haven't found a way to trigger them again for a PR.

I don't think that anyone would be using the setPage externally (and, if so, what a terrible thing to do) so I think it should probably be fine to release as a bugfix.

PR #702, opened 2024-04-19
feat: md reporter

revolunet:
In case you'd want one, a markdown reporter, useful for CI integration


josebolos:
Merci @revolunet!

Thank you so much for your contribution. As we default to text output which can easily be "transformed" into markdown is probably better for us to not add another reported that we then have to maintain.

Any users interested in a markdown reporter can still grab the code from here and use it as a custom reported.

Thanks again!

PR #718, opened 2025-05-04
Add Dependabot configuration file

aarongoldenthal:
This PR adds a `dependabot.yml` configuration file, which should restart Dependabot updates after they were previously paused, and also explicitly declares the Dependabot configuration.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    labels:
+      - 'dependencies'
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    labels:
+      - 'dependencies'


raiderss:
Great project! I'm also working on gaming assets at Eyes Store FiveM Tebex. Your code structure is impressive!

jpw (changes requested):
Hi @aarongoldenthal, many thanks for the PR.

I have made some suggestions as per our discussion. Intent is:
 - check only `production` NPM dependencies, on a weekly interval ([docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#dependency-type-allow))
 - leave GH Actions on a daily interval

aarongoldenthal:
> Hi @aarongoldenthal, many thanks for the PR.
> 
> I have made some suggestions as per our discussion. Intent is:
> 
> * check only `production` NPM dependencies, on a weekly interval ([docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#dependency-type-allow))
> * leave GH Actions on a daily interval

@jpw I got a schema error on `dependency-type`, and [digging a little further](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#allowing-specific-dependencies-to-be-updated) it needs to be within `allow`, so made that update as well. With that, it looks good to me.

jpw (approved):
Aha! Great stuff, thanks again @aarongoldenthal

Issue #709, opened 2024-06-25
Accessibility scanning with html_codesniffer as its runner fails with this error for some urls.


danghailuong1100:
Accessibility scanning with html_codesniffer as its runner fails for [this url](https://www.transizionedigitale.it/) while it is working with axe runner.
```
Error [TypeError]: Assignment to constant variable
```
This is the pa11y config in my Node project.

```
const result = await pa11y("https://www.transizionedigitale.it/", {
      chromeLaunchConfig: {
        ignoreHTTPSErrors: false,
        args: ['--no-sandbox']
      },
      wait: 3000,
      timeout: 120000,
      standard: 'WCAG2AAA',
      includeWarnings: true,
      includeNotices: true,
      runners: ['htmlcs']
});
```

## Actual behaviour
Fails with this error:
```
Error [TypeError]: Assignment to constant variable
```
## Environment

```
System:
    OS: Linux 6.5 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
    Memory: 5.20 GB / 11.64 GB
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 21.7.1 - ~/.nvm/versions/node/v21.7.1/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v21.7.1/bin/yarn
    npm: 10.5.0 - ~/.nvm/versions/node/v21.7.1/bin/npm
  pa11y: 8.0.0
```

malcomio:
This may not relate to html_codesniffer. Also, I think that a better description for this issue would be:

> TypeError when testing some URLs: Assignment to constant variable.

I've observed this for some URLs, just running pa11y directly from the command line.

I'm able to run pa11y for some URLs, but it fails for your example https://www.transizionedigitale.it/, and for https://www.neso.energy/ :

```
pa11y https://www.neso.energy/     

Welcome to Pa11y

 > Running Pa11y on URL https://www.neso.energy/

Error: TypeError: Assignment to constant variable.
    at h (evaluate at injectRunners (/Users/mayoung/.nvm/versions/node/v20.18.1/lib/node_modules/pa11y/lib/pa11y.js:300:20), <anonymous>:32:289113)
[...]

PR #698, opened 2024-03-18
[WIP, reproduced as failing test] Support pages containing `<input autocomplete="username">`

danyalaytekin:
Adding a failing test to draft PR.

- Will fix #691 
- Renames #373's test objects, moves its commentary into code, and inverts and broadens its assertion to simply check for success. Unfortunately I can't seem to make the `git mv` hold here 🤔 so the tracking has been lost.

PR #734, opened 2025-07-09
Switch to fork of HTML_CodeSniffer

kant2002:
which fix https://github.com/pa11y/pa11y/issues/709


danyalaytekin:
Hi @kant2002, thank you for your contribution. We've been [discussing](https://github.com/pa11y/pa11y/pull/698#issuecomment-2005578428) taking the step of a soft fork, so you're on the right track here, but we would need to maintain the fork itself under the `pa11y/` org.

PR #848, opened 2026-08-26
Resolving the Symlink Vulnerability of extract-zip

sangitamane:
Upgrading from Puppeteer 25.8.0 to 25.9.0 fixes the security issue:
 `extract-zip unvalidated symlink path traversal`
The @puppeteer/browsers Upgrade: The Puppeteer ecosystem officially removed the old, frozen extract-zip module from its internal tooling


josebolos (approved):
Great! We finally can get rid of that alert.

PR #852, opened 2026-09-18
feat(axe): keep the failure summary and check data on violations

darrenbritton:
axe computes a per-node `failureSummary` ("Fix any of the following: …") and, for rules like `color-contrast`, check `data` with the measured and expected ratios. The runner kept neither on violations, so downstream (the platform's issue text, the Playwright package's local failures) can only name the rule, not what is wrong with the element.

This adds `runnerExtras.failureSummary` to every issue that has one, and `runnerExtras.axeData` to violations whose matching check carries data, the same field incompletes already have. Both are added only when present, so existing consumers and tests are unaffected. Codes, messages and selectors are untouched, so hashing is unchanged.

Unit test added; `npm run test-unit` 509 passing, lint 0 errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## 9.2.0 (2026-05-25)
+
+First release of the `pa11y-unlimited` fork.
+
+* Remove the limit on HTML context in error reports so the full surrounding markup is preserved.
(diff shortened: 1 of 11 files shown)


darrenbritton:
Opened against the wrong repository by mistake; this change targets a downstream fork. Apologies for the noise.

